PR #6, opened 2020-05-01
Refactored source catalog and getFlowByActivity

WesIngwersen:
I had [identified the need to identify fba datasets with sector-like activity schemas
](https://epalca.atlassian.net/browse/FLOWSA-97). I determined it was best to add this information into the source catalog, but the best way to do this was a significant refactor of this yaml. This commit does that.  As a result the getFlowByActivity function in init was greatly simplified, but now is restricted to only return flows from a named datasource, where datasource is mandatory.


WesIngwersen:
@catherinebirney I just updated it to resolve a conflict because you had just changed the name of the CBECS source in response to another comment.

PR #26, opened 2020-08-28
Update scc crosswalk

bl-young:


WesIngwersen:
@bl-young  @cchiq any comments on the actual changes made?

PR #41, opened 2020-12-03
Refactor flowbysector

catherinebirney:
Functions refactored to require Activity/SectorProducedBy and Activity/SectorConsumedBy rather than single 'Activity' and 'Sector' columns. Refactored functions do not change the output of NEI or water FBS.


flowsa/data/flowbysectormethods/Land_national_2012.yaml
```diff
@@ -1,11 +1,83 @@
 %YAML 1.2
-
+# 'EIA_CBECS_Land' and 'EIA_MECS_Land' must be listed prior to 'USDA_ERS_MLU' for FBS creation
```
bl-young, line 2: Out of curiosity, why is that?

bl-young (approved):

flowsa/data/flowbysectormethods/Land_national_2012.yaml
catherinebirney: @bl-young. In order to allocate urban land use to sectors, we need to subtract land area used for commercial and manufacturing buildings from the MLU value. So the CBECS and MECS FBS dfs are added to 'fbs_list', which the MLU allocation function calls on.

PR #43, opened 2020-12-14
Refactor init to use processed data mgmt from new esupy library

WesIngwersen:
This simplifies greatly the init and adds processed data handling coming from the new esupy library.

Needs to be independently tested with pulling and trying an install and make sure it works


catherinebirney (changes requested):
@WesIngwersen a few comments: 1) I get an error message for getFlowBySector - the function is missing the argument 'paths'. 2) Do you want the script 'processed_data_mgmt.py' in flowsa? The functions are modified versions of those in esupy. 3) When I run getFlowByActivity, I get a log error saying that the file isn't found on my local or remote server, but the function does actually work

WesIngwersen:
@catherinebirney @bl-young  I'm going to work on this more because aside from Catherine's comments I found another couple issues due to the migration of the code to esupy in the middle of the refactor. We can leave the PR open to preserve the discussion and i will mention you when i push fixes, some of which will be in esupy

WesIngwersen:
@catherinebirney please reinstall esupy, pull these updates and try again, thanks

catherinebirney:
@WesIngwersen No more issues running the code

WesIngwersen:
@catherinebirney @bl-young  This is ready for review but the [esupy PR](https://github.com/USEPA/esupy/pull/1) should be reviewed and pulled in first because this relies on the code in the dataplan_refac  branch. Then conflicts will need to be reconciled. @catherinebirney we could try to do this together

flowsa/common.py
```diff
@@ -36,10 +40,17 @@
...
 flowbysectormethodpath = datapath + 'flowbysectormethods/'
 flowbysectoractivitysetspath = datapath + 'flowbysectoractivitysets/'
-fbaoutputpath = outputpath + 'FlowByActivity/'
 fbsoutputpath = outputpath + 'FlowBySector/'
 
-local_storage_path = appdirs.user_data_dir()
+paths = Paths
+paths.local_path = os.path.realpath(paths.local_path + "/flowsa")
+
+pkg = pkg_resources.get_distribution("flowsa")
```
bl-young, line 48: This is writing out for me as 0.0.0.1 (which means it appears to have an extra 0?)

flowsa/__init__.py
```diff
@@ -48,20 +42,8 @@ def getFlowBySector(methodname):
...
-        # if parquet does not exist in local repo, read file from Data Commons
-        try:
-            log.info(methodname + ' parquet not found in local repo, loading from Data Commons')
-            fbs = pd.read_parquet('https://edap-ord-data-commons.s3.amazonaws.com/flowsa/FlowBySector/' +
-                                  methodname + ".parquet")
-        except FileNotFoundError:
-            log.error("No parquet file found for datasource " + methodname + " in flowsa or Data Commons")
-
+    fbs_file = "FlowBySector/" + methodname + ".parquet"
+    fbs = load_preprocessed_output(fbs_file, paths)
```
bl-young, line 46: Looks like you haven't yet gotten to FBS, as need to assign fbs_meta to pass to load_preprocessed_output

catherinebirney:
I get an error when I run 'import flowsa'
pkg_resources.DistributionNotFound: The 'flowsa' distribution was not found and is required by the application

I do not get this error in the master branch. I can look into it more tomorrow

@WesIngwersen I agree that it we should go through the conflicts together

WesIngwersen:
Thanks. I added that to flows.common and import it into init so that would make since it woukd throw an error on import. Yes we can troubleshoot it together

flowsa/common.py
catherinebirney: fxn output for me is 0.0.1, could we preface with a 'v' to clarify it is the version?

catherinebirney:
Thanks - now able to run

getFlowByActivity works well
I'm having the same problem as Ben with getFlowBySector, which is missing the meta assignment

flowsa/__init__.py
WesIngwersen: Yes, I've only made the changes for fba's as  a starting place

flowsa/common.py
WesIngwersen: For me it was also just 0.0.1 -- I'm interested in seeing that. Yes I  will add the 'v' which is consistent with the schema. That is done in esupy

WesIngwersen:
This is now updated to work for flowBySector as well.   The change to flowbyactivity calls (See Jira FLOWSA-317 not yet assigned) as well as issue #59 should be addressed before pulling in

flowsa/common.py
```diff
@@ -43,7 +43,7 @@
 fbsoutputpath = outputpath + 'FlowBySector/'
 
 paths = Paths
-paths.local_path = os.path.realpath(paths.local_path + "/flowsa")
+paths_local_path = os.path.realpath(paths.local_path + "/flowsa")
```
bl-young: I believe local_path is a variable within the class Paths, so this needs to remain paths.local_path (I think)

catherinebirney: yes, my mistake, I changed it back

WesIngwersen:
@catherinebirney Still having merging issues?

catherinebirney:
> @catherinebirney Still having merging issues?

No, merging the branches in pycharm is straightforward. There are significant differences in flowbyactivity.py that I need to reconcile, so I'm working through the error messages. Will hopefully have FBA and FBS working in the next few hours and then I'll commit.

catherinebirney:
@bl-young @WesIngwersen @MoLi7 @melissagqc 
Changes are merged with Master

flowsa/common.py
bl-young: In reviewing this step for lciaformatter, I'm wondering whether how a package gets installed could be affecting this. I found I needed to update my egg because back when I originally installed, setup.py indicated a much earlier version. Do you all do an editable install, and if so, do you manually update your egg? (i used python setup.py bdist_egg to do so).

I believe this explains my original version being 0.0.0.1.
Since Catherine recently updated the version number in flowsa, I wonder if in pycharm for you all your egg will update automatically or if it will still mark output files as v0.0.1
@WesIngwersen @catherinebirney

catherinebirney: I'm having trouble with this step too. setup.py is updated to reflect version v0.0.2, but any FBA or FBS output parquet still reflects version v0.0.1, which is the version indicated in my egg

bl-young: I wonder if we could set a global variable in each package, which would get used within setup.py, and could also be passed to the meta definition for esupy. That way you change it in one spot and wouldn't have to worry about re-creating the egg each time. I suppose, fortunately, this is only an issue for developers and not users.

PR #46, opened 2021-01-12
Land fbs updates

catherinebirney:
1) updates to Land FBS yaml and methodology for land allocation
2) code added to ensure units of different dfs compatible when combined (focused on water FBS)
3) moved fxns to subset dfs by geography to immediately after loading an FBA to speed up fxns when loading BLS QCEW data (the BLS FBA's now have county data which makes the files large and slow to process)


WesIngwersen:
@catherinebirney Did some of these changes help to allocate the remaining 10% of flows?

catherinebirney:
@WesIngwersen Yes! And I've pinpointed the remaining dropping FlowAmounts to Mining. Just have to work out exactly what causes data to be dropped after loading the BLM_PLS FBA

bl-young:
@catherinebirney I will try to rerun NEI data at some point but dont feel like you have to wait on me before pulling this in, unless you have specific concerns.

PR #48, opened 2021-01-22
add 'MetaSources' column to FBS

catherinebirney:
Changes:
1. Adds MetaSources to FBS output
2. Adds fxn harmonize_FBS_columns(), which ensures there aren't multiple rows of data for a SectorProducedBy/SectorConsumedBy combo where the only differences in rows are FIPS year/MetaSource/MeasureofSpread


bl-young:
@catherinebirney so you successfully built CAP_HAP? did you try any others?

catherinebirney:
@bl-young GRDREL_national_2017 runs. I get an error with TRI_DMR_national_2017:
FileNotFoundError: [Errno 2] No such file or directory: '...Python/Python37/lib/site-packages/stewi/data/DMR/2_digit_SIC.csv'

bl-young:
Ok probably the same issue as yesterday. I will adjust this line to make sure the entire data file is included in the install:
https://github.com/USEPA/standardizedinventories/blob/master/setup.py#L34

catherinebirney:
@bl-young There is a problem with the latest commit, I get an error message when using the command line to install stewi:
ValueError: path 'data/' cannot end with '/'

bl-young:
ok sorry try now

catherinebirney:
Thanks - fixed the issue. I can now run TRI_DMR

bl-young (approved):

bl-young:
ok did some testing myself on the data that comes out of StEWI and looks good to me.

bl-young:
actually, @catherinebirney, given what you did in harmonize_fbs_columns, it's now safe for me to assign the datasource name individually from stewicombo, correct, and they should get combined. Let me check on that.

catherinebirney:
> actually, @catherinebirney, given what you did in harmonize_fbs_columns, it's now safe for me to assign the datasource name individually from stewicombo, correct, and they should get combined. Let me check on that.

Yes, the datasource names should get combined

bl-young:
will data quality get aggregated and averaged correctly?

catherinebirney:
Data quality scores are calculated with a weighted average through the aggregator fxn
https://github.com/USEPA/flowsa/blob/master/flowsa/flowbyfunctions.py#L180
which is run after harmonize_fbs_columns. I didn't do any data testing (which I should), but the scores should be averaged correctly. I'll add a jira task to test the results

bl-young:
Perfect, yeah I just tested in the TRI/DMR data and the data quality was the same as before for all flows. I misunderstood your function.

I will add that this can manifest as some records with "TRI, DMR" and others with "DMR, TRI". I dont know if there is an easy solution to that.

catherinebirney:
> I will add that this can manifest as some records with "TRI, DMR" and others with "DMR, TRI". I dont know if there is an easy solution to that.

Ah okay, good catch! I wonder if sorting the df by MetaSources prior to joining MetaSources names would fix that. I'll check.

catherinebirney:
@bl-young sorting the df by metasources prior to concating the metasource strings appears to have fixed the issue - although only tested with TRI_DMR. If you're happy with the changes I'll go ahead and merge

bl-young (approved):

PR #51, opened 2021-02-02
Mastercrosswalk refactor

catherinebirney:
1. incorporates revised useeior mastercrosswalk, which includes the household and government transportation sectors
2. adds a sector check - if sectors are not naics 2012, checks if naics exist for another year (2007 and 2017), if so, replaces sectors with naics 2012 code. If it is not a 1:1 mapping between years, equally allocates the FlowAmount to all related naics 2012


catherinebirney:
@bl-young There are a few instances where a sector code does not belong to any year. In TRI_DMR_national_2017, '541710' cannot be replaced. The code prints out any sectors that are not NAICS 2012.

bl-young:
> There are a few instances where a sector code does not belong to any year. In TRI_DMR_national_2017, '541710' cannot be replaced. The code prints out any sectors that are not NAICS 2012.

What happens to these, I assume they get dropped? Or are they maintained?

flowsa/__init__.py
```diff
@@ -7,73 +7,81 @@
...
 
-
-def getFlowByActivity(flowclass, years, datasource, geographic_level='all'):
+def getFlowByActivity(flowclass, years, datasource, geographic_level='all', file_location='local'):
     """
     Retrieves stored data in the FlowByActivity format
     :param flowclass: list, a list of`Class' of the flow. required. E.g. ['Water'] or
      ['Land', 'Other']
-    :param year: list, a list of years [2015], or [2010,2011,2012]
+    :param year: list, a list of years [2015], or [2010,20flowclass, years, datasource, geographic_level='all', file_location='local'11,2012]
```
bl-young, line 20: typo?

flowsa/__init__.py
```diff
@@ -7,73 +7,81 @@
...
     :param datasource: str, the code of the datasource.
-    :param geographic_level: default set to 'all', which will load all geographic scales in the FlowByActivity, can \
```
bl-young, line 22: I also wonder if you meant to delete this or not

bl-young (approved):
outside of those typos looks good to me. I will test the updated data from stewi in master

flowsa/__init__.py
catherinebirney: Ha! Thank you, corrected this typo (and the next)

catherinebirney:
> > There are a few instances where a sector code does not belong to any year. In TRI_DMR_national_2017, '541710' cannot be replaced. The code prints out any sectors that are not NAICS 2012.
> 
> What happens to these, I assume they get dropped? Or are they maintained?

They are maintained

PR #63, opened 2021-02-25
update the naics crosswalk

catherinebirney:
naics crosswalk changes due to updating useeior mastercrosswalk


bl-young:
This just brings this branch current with master? Yeah that's fine. I've only made a few minor changes to SCC crosswalk here so far

bl-young (approved):

catherinebirney:
yes - I had updated the employment_refactor branch with the revised naics crosswalk from useeior and only now pulled it into master

PR #71, opened 2021-03-17
refactor fbs

catherinebirney:
- update the NEI MECS allocation to allocate to NAICS6 with BLS QCEW Employment data
- reorganize/restructure the fxns in FBS


flowsa/flowbyactivity.py
```diff
@@ -143,6 +144,8 @@ def main(**kwargs):
 
     # assign yaml parameters (common.py fxn)
     config = load_sourceconfig(kwargs['source'])
+    # update the local config with today's date
+    config['date_generated']= pd.to_datetime('today').strftime('%Y-%m-%d')
```
bl-young, line 148: @catherinebirney I recognize that no bib is made during the FBA generation process, but I figured best to update the date here locally too after its pulled in, in case in the future the bib gets generated automatically.

catherinebirney: Ah ok, good call

bl-young:
@catherinebirney I believe we are ending up with just a handful of negative allocations from the EIA_MECS data:

  | Flowable | Activity | MetaSources | FlowAmount
-- | -- | -- | -- | --
4657 | Particulate matter, ≤ 2.5μm | 331490 | EPA_NEI_Nonpoint | (9,773)
4654 | Particulate matter, ≤ 2.5μm | 331420 | EPA_NEI_Nonpoint | (12,690)

Maybe this results from the suppressed data? Otherwise this is looking good

catherinebirney:
Ok, thanks @bl-young, I'll look into the negative numbers

PR #73, opened 2021-03-19
updates to fbs yaml formatting for stewi 

catherinebirney:
- modify formatting of fbs yamls that call on stewi (fbs_outside_flowsa)
- modify fxn arguments when loading fbs in flowbysector.py

@bl-young I implemented the changes we discussed and tested for cap_hap, crhw, grdrel, and tri_drm


flowsa/data_source_scripts/stewiFBS.py
```diff
@@ -36,18 +36,24 @@ def stewicombo_to_sector(yaml_items):
     import stewicombo
     from flowsa.data_source_scripts.EPA_NEI import drop_GHGs
 
-    NAICS_level_value = sector_level_key[yaml_items['NAICS_level']]
+    # determine if fxns specified in FBS method yaml
+    if yaml_load['functions'] != None:
```
bl-young, line 40: I think this should be is None

catherinebirney: oh, I think it should be 
`if 'functions' not in yaml_load:`

catherinebirney: yes, the above syntax works if 'function' is not in the fbs yaml. 
`if yaml_load['functions'] is None:`
throws an error

bl-young (approved):

PR #97, opened 2021-08-17
update master branch with unit refrac branch

catherinebirney:
- FBA units are unchanged on import
- all FBA units are standardized when first loaded to create flowbysector
- includes changes from validation branch, new log file for validation only
- updates to mapping - relies on esupy.mapping


catherinebirney:
@bl-young I don't have 'materialflowlist' installed, so not able to check the impact the changes have on stewi data

bl-young:
should we do another release of esupy for mapping.py?

catherinebirney:
> should we do another release of esupy for mapping.py?

Yes, let's

bl-young:
@catherinebirney I made some additional changes upon testing the stewi datasets. otherwise things seem to be working fine

bl-young (approved):
i think ok to pull in

catherinebirney:
Well, now that you've approved, I made one big change. See here https://github.com/USEPA/flowsa/commit/96a20311b8697d15859a2a07132d7ca8fdc3d532

bl-young:
@catherinebirney ok I will try to test again tomorrow using the esupy metadata branch. I will need to confirm that https://github.com/USEPA/esupy/pull/13 does not impact things in StEWI

bl-young:
In init.py, will the subdirectory dict also apply to FBA? or do you expect it to be different between FBA/FBS? Or should this be a common variable.

Also, line 79 & 33, now that esupy is released can these import statements be moved up?

catherinebirney:
Ah, good catch. Right now all log files regardless of being FBA or FBS are stored in the same Log folder. However, it would probably be a good idea to have seperate log folders to make information easier to find. So I do not think I want the dict to be a common variable. But, we could move the dictionaries to 'common.py' rather than defining them in the init file. 

And yes, we should try to move the esupy import statement to a top level - just have to make sure we don't get circular logic issues

bl-young (approved):

PR #104, opened 2021-09-03
Validation

catherinebirney:
- modify FBS yaml to drop unnecessary parameters
- moved some functions to a new .py file
- renamed "mapping.py" to "sectormapping.py" to clarify sector mapping from fedefelm mapping
- add functions seeAvailableModels() for use in the examples folder so users know what to input
- modifications to some FBA yamls to correct years of available data
- updates to Census PEP and Census CBP FBA because there were errors in creating FBAs for certain years (overloaded the API and some years of data were not available)
- updates pandas and xlrd requirements to address any possible errors with loading xlsx and xls files
- add explanations throught water and land FBS for why FlowAmounts are modified
- incorporate more validation checks within FBS.py


catherinebirney:
@bl-young Trying new method of a "develop" branch before merging into master for new release

examples/get_flows_by_activity.py
```diff
@@ -12,19 +12,29 @@
     :param geographic_level: str, a geographic level of the data.
     Optional. E.g. 'national', 'state', 'county'.
     :return: a pandas DataFrame in FlowByActivity format
+
+If you encounter the error:
+pkg_resources.DistributionNotFound: The 'flowsa' distribution was
+not found and is required by the application
+
+See the flowsa wiki:
+https://github.com/USEPA/flowsa/wiki/GitHub-Contributors#troubleshooting
```
bl-young, line 21: this link is not active

catherinebirney: thanks, fixed!

bl-young:
This may not be entirely relevant to this PR, but did we determine if the pkg_resources.get_distribution function works correctly for users that install from pip? It looks like that function is only used to get the name 'flowsa' to assign it in the metadata. So is that even necessary?

catherinebirney:
@bl-young , no never checked. But you're right, it is just used to get 'flowsa'. I will just set pkg = 'flowsa' and then I can remove the comment about encountering the error

bl-young:
I think seeAvailableFlowByModels is a good function. Perhaps it should go in init?

bl-young (approved):

catherinebirney:
> I think seeAvailableFlowByModels is a good function. Perhaps it should go in init?

@bl-young good call, will move to init

PR #114, opened 2021-09-14
Support testing with github actions

andychase:
These set of changes support automatically running integration tests in the repository.

* Examples moved to flowsa/ to fix import issues
* Common Api keys now prints the key missing and uses os-independant coding for paths
* Requirements.txt is removed and the setup.py file is used for requirements
* Github actions script is added
  * This script requires a new "secret" USDS_KEY in order to work to be added to the github project. This should be set to USDA_Quickstats_API_KEY
* pytest.ini added which sets an ignore directory and enables doctest
* Examples no longer run on load, which enables scanning by the test runner
* A test_examples file is added which tests the examples


catherinebirney:
@andychase Thanks, Andy. Will spend some time looking into these changes. We do have to keep the requirements.txt per EPA guidelines for publishing a python package.

andychase:
You should go the other direction then and make the setup.py support whats
in the requirements.txt

https://stackoverflow.com/a/16624700

I'm still working on my full comments

PR #126, opened 2021-11-01
Fix for the 1st mearge errors on Git actions

melissagqc:
@catherinebirney  I am pulling the changes back to my branch.


catherinebirney:
@melissagqc great, thanks for addressing the import errors

PR #130, opened 2021-11-09
update develop with AC and DG comments

catherinebirney:
1. Use a .env file to hold the API_Keys instead of .txt files (add .env to .gitignore), add an example .env file to 'examples' directory
2. Condense functions (loading crosswalks and yaml dictionaries)
3. new function "equal_allocation" that is called on in the "direct_allocation" method which will check if an Activity is being allocated to more than one sector and if so, will equally allocate the Activity FlowAmount to all sectors
4. new script "schema.py" that holds all of the FBA and FBS schema dictionaries
5. Adds a memory cache using 'joblib' so can quickly regenerate FBAs if url has already been called
6. new function str2bool() used to ensure a bool always returned when there is an input for 'download_FBAs_if_missing' when generating FBS
7. reduce use of **kwargs parameters throughout to make more clear when the function inputs/requirements are
8.  Add fxn docstrings throughout
9.  Rename estimate_suppressed_data() to equally_allocate_parent_to_child_naics() to clarify what the fxn does
10.  use git describe to return package version number for use in meta and filename instead of manually updating a version number in common.py
11. two new package requirements joblib (for url memory cache) and python-dotenv (Reads .env files)


bl-young:
@catherinebirney  I'm reviewing now - have you attempted to generate some of the FBAs on this new branch? (like NEI for instance)

catherinebirney:
@bl-young Yes, I generated all FBS, which generated all the necessary FBAs

flowsa/data_source_scripts/EPA_NEI.py
```diff
@@ -108,12 +108,12 @@ def epa_nei_global_parse(dataframe_list, args):
     df = df[~df['Location'].str[-3:].isin(excluded_fips2)]
 
     # drop all other columns
-    df.drop(df.columns.difference(columns=['FlowName',
-                                           'FlowAmount',
-                                           'ActivityProducedBy',
-                                           'Location',
-                                           'Unit',
-                                           'Description']), inplace=True)
+    df.drop(df.columns.difference(['FlowName',
```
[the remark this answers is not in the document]
catherinebirney: I got an error when I switched it to "column" so I switched it back - I  meant to look into how to correctly change it

bl-young:
@catherinebirney finding a potential issue with some of the stewi FBS let me look a bit closer

flowsa/settings.py
```diff
@@ -84,8 +83,23 @@
 vLog.addHandler(vLog_fh)
 vLogDetailed.addHandler(vLog_fh)
 
+
+def return_pkg_version():
```
bl-young, line 87: If we find this works well we could consider moving to esupy similar to `get_git_hash()`

bl-young (approved):

flowsa/settings.py
catherinebirney: @bl-young Yes, agreed. I do think it is a better method than manually updating a parameter in a .py file

catherinebirney:
> @catherinebirney finding a potential issue with some of the stewi FBS let me look a bit closer
> 
> ok I think this is because I'm using more up to date facility data from TRI ([which is wrong](https://github.com/USEPA/standardizedinventories/issues/97)). So if its working for you probably ok.

Ok, good - I don't have any issues generating stewi FBS from scratch

Issue #148, opened 2021-12-08
Underutilization of `.yaml`s


matthewlchambers:
Raising this as an issue now for awareness and discussion. I think we are currently hardcoding too many things in the data source script `.py` files and not making enough use of the `.yaml` files. For example, information on how to modify the NEI data url for different years, or information on how columns should be renamed (again, I'm thinking about the NEI data source in particular at the moment) and how the renaming depends on the year, could be included as dicts in the `EPA_NEI.yaml` files.

One advantage of offloading this information to the `.yaml` files is that it should simplify the process of adding new data years (e.g. the 2020 NEI, or 2018 and 2019 EQUATES data).

I'm working on the FBA methods and scripts for the EQUATES data right now, and I'll link to them once I'm finished, by way of an example of what I'm thinking.

matthewlchambers:
To this point, for the EQUATES data it would be extrememly helpful if I could pass the `.yaml` information to the calling and parsing functions. I see two possible ways to do this: 1.) Combine the `config` dict into `args` before passing it; 2.) Add `config` as an arg to all the calling and parsing functions. 1 is easier, 2 is more consistent with how other `source.py` function work (and we should probably be moving away from folding things into `args`). I lean toward 2, but I'm willing to go through and implement whichever option is preferred.

catherinebirney:
I agree with method 2

bl-young:
@matthewlchambers your option 2 above is implemented in #150 correct?

Once that gets pulled in I can refactor the NEI, EIA_MECS, and ghgrp yamls to consolidate that information.

matthewlchambers:
@bl-young Yes, that's correct!

bl-young:
Updates have been made for NEI, GHGI, and EIA_MECS. We can consider others as we need to update them. Should we close this issue for now?

matthewlchambers:
> Updates have been made for NEI, GHGI, and EIA_MECS. We can consider others as we need to update them. Should we close this issue for now?

Sorry, didn't see your comment until now. Yes, I think it can be closed.

PR #154, opened 2021-12-17
Recursive `.yaml` inheritance

matthewlchambers:
This PR implements a new optional entry in `source.yaml` files. If a `source.yaml` file contains the key `inherits_from`, the value associated with this key is interpreted as the name of another `.yaml` file in the same folder which should be read in and combined with the original `source.yaml` file to produce the `config` dictionary.

Merging the dictionaries is done such that if `parent.yaml` and `child.yaml` contain the same key, the value from `child.yaml` will be the one used in `config`. A warning is logged in this situation so the user will be aware.

Additionally, inheritance is allowed to be recursive: `parent.yaml` may have the entry` inherits_from: grandparent.yaml`, and so on.

This should be helpful for sources like EQUATES and NEI which would otherwise have a lot of duplicated information across `.yaml` files. Now, for example, all common information for EQUATES can be held in `EQUATES_Common.yaml`, with nonpoint source specific information in `EQUATES_Nonpoint.yaml`. In this case, `EQUATES_Nonpoint.yaml` contains the entry `inherits_from: EQUATES_Common`.

### DIFF
--- a/flowsa/common.py
+++ b/flowsa/common.py
@@ -89,12 +89,31 @@ def load_yaml_dict(filename):
     Load the information in 'source_catalog.yaml'
     :return: dictionary containing all information in source_catalog.yaml
     """
-    if filename == 'source_catalog':
-        yaml_load = f'{datapath}source_catalog.yaml'
-    else:
-        yaml_load = sourceconfigpath + filename + '.yaml'
-    with open(yaml_load, 'r') as f:
+    folder = datapath if filename == 'source_catalog' else sourceconfigpath
+    yaml_path = folder + filename + '.yaml'
+
+    with open(yaml_path, 'r') as f:
         config = yaml.safe_load(f)
+
+    # Allow for .yaml files to recursively inherit other .yaml files. Keys in
+    # children will overwrite the same key from a parent.
+    inherits = config.get('inherits_from')
+    while inherits:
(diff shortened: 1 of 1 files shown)


bl-young:
@matthewlchambers I like this idea - I had just started updating the NEI `.yaml` files re: #148 and was thinking about ways to avoid having to repeat the data dictionaries in each. (I was leaning towards the opposite approach of combining them into a single `.yaml` but this works just as well and will likely be more flexible moving forward)

catherinebirney:
I also like this idea - it will also make it more clear what the differences are between yamls when we have more than one method of generating an FBS

bl-young:
> when we have more than one method of generating an FBS

I believe this currenlty only impacts FBA.

FBS yaml files are read [here](https://github.com/USEPA/flowsa/blob/develop/flowsa/flowbysector.py#L71-L83), but really that should probably use the existing function in common so could work for FBS as well.

catherinebirney:
@bl-young Ah yes, got ahead of myself - yes, the FBS should be modified to utilize @matthewlchambers changes

matthewlchambers:
> @bl-young Ah yes, got ahead of myself - yes, the FBS should be modified to utilize @matthewlchambers changes

See #157, which implements this modification for FBS.

PR #157, opened 2021-12-21
Make FBS use load_yaml_dict from common

matthewlchambers:
This PR expands the `common.py` function `load_yaml_dict()` to let it read in flow by sector `.yaml`s as well as flow by activity `.yaml`s. The `flowbysector.py` function `load_method()` is removed and replaced with `load_yaml_dict()` from `common.py`.

### DIFF
--- a/flowsa/common.py
+++ b/flowsa/common.py
@@ -89,7 +89,7 @@ def load_yaml_dict(filename, flowbytype=None):
     Load the information in 'source_catalog.yaml'
     :return: dictionary containing all information in source_catalog.yaml
     """
-    if filename == 'source_config':
+    if filename == 'source_catalog':
         folder = datapath
     else:
         if flowbytype == 'FBA':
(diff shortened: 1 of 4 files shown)


flowsa/common.py
```diff
@@ -84,16 +84,27 @@ def load_crosswalk(crosswalk_name):
...
 
 
-def load_yaml_dict(filename):
+def load_yaml_dict(filename, flowbytype=None):
     """
     Load the information in 'source_catalog.yaml'
     :return: dictionary containing all information in source_catalog.yaml
     """
-    folder = datapath if filename == 'source_catalog' else sourceconfigpath
+    if filename == 'source_config':
```
catherinebirney, line 92: @matthewlchambers Line 92 sound read `if filename == 'source_catalog:`

matthewlchambers: Fixed! Glad you caught that!

PR #160, opened 2021-12-23
Equates FBA

matthewlchambers:
This PR provides `source.yaml` and `source.py` files for the EPA EQUATES project data, along with the following changes to other flowsa code:

1. `flowbyactivity.call_urls()` modified to allow use of the new `confirm_gdrive` capability of `make_url_request()`
2. `flowbyfunctions.assign_fips_location_system()` modified to add column to dataframe in a way that doesn't cause an error if the dataframe happens to be empty


catherinebirney (approved):

bl-young:
@matthewlchambers I see there's a fair bit of duplication from the NEI file? Was that intentional? Should just host that code in one place? Or do you anticipate modifying it for equates data?

matthewlchambers:
> @matthewlchambers I see there's a fair bit of duplication from the NEI file? Was that intentional? Should just host that code in one place? Or do you anticipate modifying it for equates data?

I definitely should have deleted that portion of the code. I started from `EPA_NEI.py` because there were some similarities between the NEI and EQUATES data, but forgot to delete the functions I didn't need. Thanks for catching that!

PR #163, opened 2022-01-03
add data visualization

catherinebirney:
- plot to compare differences between methodologies for same flow
 - plot to visualize flows for subset of sectors
 - adds missing naics6 sector names to 2012 crosswalk
 - seaborn and matplotlib additional required packages


catherinebirney:
@bl-young requesting your review more as an FYI

bl-young:
looks nice. I was able to run the example script but didn't try to modify or create my own

bl-young (approved):

PR #174, opened 2022-01-27
Transportation Satellite Account methods and scripts, draft 1

matthewlchambers:
Right now, there's an issue where the data frame is successfully parsed, but then somehow the FBA is empty. I may need some help figuring out what's going on here.

This pull request only adds the `BTS_TSA.yaml` and `BTS_TSA.py` files.

### DIFF
--- a/flowsa/data_source_scripts/BTS_TSA.py
+++ b/flowsa/data_source_scripts/BTS_TSA.py
@@ -1,0 +1,29 @@
+# BEA.py (flowsa)
+# !/usr/bin/env python3
+# coding=utf-8
+
+'''
+Pulls BTS Transportation Satellite Account (TSA) data
+'''
+from io import BytesIO
+import pandas as pd
+from flowsa.common import US_FIPS, fbs_activity_fields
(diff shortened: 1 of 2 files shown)


flowsa/data_source_scripts/BTS_TSA.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def tsa_call(*, resp, year, **_):
+    '''
+    Convert response to pandas dataframe
+    :param resp: response from url call
+    :return: pandas dataframe of original source data
+    '''
+    df = pd.read_csv(BytesIO(resp.content))
+    df = df[df.year == year]
```
bl-young, line 24: @matthewlchambers  I think year is a string in the function, but in your df is an int

matthewlchambers: Dang. Good eye! That was indeed the problem. In testing I was calling the `tsa_call()` function separately, with `year` as an int. Hence why it appeared that the problem was downstream from here.

PR #176, opened 2022-01-27
Fixes #175

### LINKED ISSUE #175: Transportation Satellite Account FBA not generating properly
I've verified that the TSA data frame is correctly parsed by the `tsa_parse()` function from `BTS_TSA.py`, but then the FBA is empty. So somewhere between the data frame being parsed and the final FBA being written, all the rows are dropped. I've spent some time on it, but I may need some help figuring out this issue.

matthewlchambers:
@bl-young found the problem. This PR fixes it.


catherinebirney:
Thanks @matthewlchambers @bl-young , I got ahead of myself by merging in the other pr!

PR #181, opened 2022-03-02
enable multi sector levels within FBS

catherinebirney:
- Enables additional, less aggregated, sector levels in final FBS dataset

- When subsetting data based on sector levels, FLOWSA functions generally take the string length of a sector. This method causes issues when the sector is a household or government sector code, where the string length can deviate from the sector level. The revised code subsets dataframes using the sector length crosswalk, which correctly categorizes household/government codes. 

- Updates the NAICS crosswalk to include NAICS 7


scripts/FlowByActivity_Crosswalks/write_Crosswalk_USDA_CoA_Livestock.py
```diff
@@ -33,8 +33,8 @@ def assign_naics(df):
     # dual-purpose cattle ranching and farming: 11213
     df.loc[df['Activity'] == 'CATTLE, (EXCL COWS)', 'Sector'] = '112130A'
     df.loc[df['Activity'] == 'CATTLE, COWS', 'Sector'] = '112130B'
-    df.loc[df['Activity'] == 'CATTLE, COWS, BEEF', 'Sector'] = '112130B1'
-    df.loc[df['Activity'] == 'CATTLE, COWS, MILK', 'Sector'] = '112130B2'
+    # df.loc[df['Activity'] == 'CATTLE, COWS, BEEF', 'Sector'] = '112130B1'
+    # df.loc[df['Activity'] == 'CATTLE, COWS, MILK', 'Sector'] = '112130B2'
```
bl-young, line 37: are these supposed to be removed?

flowsa/methods/flowbysectormethods/Water_state_2015_m1.yaml
```diff
@@ -1,4 +1,5 @@
 target_sector_level: NAICS_4
+target_subset_sector_level: {NAICS_6: ['1133', '1125']}
```
bl-young, line 2: Are these true targets here or just sample? A small comment might help to clarify why those are being targeted.

scripts/FlowByActivity_Crosswalks/write_Crosswalk_USDA_CoA_Livestock.py
catherinebirney: > I wonder if we can re-think this file a bit. Right now because each section is numbered making a small change results in lots of line updates so its hard to track. I suspect a lot of these change in `fbs_yaml_refac`
> 
> Bullets might be better as in most cases the order doesn't matter. Maybe we could lead with the required fields followed by the optional fields and/or use a table.

I like this idea. I see I updated this file prematurely with the revised parameters in the fbs_refac_branch

catherinebirney:
> I wonder if we can re-think this file a bit. Right now because each section is numbered making a small change results in lots of line updates so its hard to track. I suspect a lot of these change in `fbs_yaml_refac`
> 
> Bullets might be better as in most cases the order doesn't matter. Maybe we could lead with the required fields followed by the optional fields and/or use a table.

I like this idea. I see I updated this file prematurely with the revised parameters in the fbs_refac_branch

bl-young:
> > I wonder if we can re-think this file a bit. Right now because each section is numbered making a small change results in lots of line updates so its hard to track. I suspect a lot of these change in `fbs_yaml_refac`
> > Bullets might be better as in most cases the order doesn't matter. Maybe we could lead with the required fields followed by the optional fields and/or use a table.
> 
> I like this idea. I see I updated this file prematurely with the revised parameters in the fbs_refac_branch

Sorry I submitted before I finished my thought - I think I was saying that it's going to undergo some changes in that branch anyway so probably makes sense to make further changes there instead.

bl-young:
@catherinebirney Should we bring in #185 first so as to confirm that this doesn't break any existing FBS? I'm not sure how to test this new functionality as we don't have any methods yet with this implemented

catherinebirney:
@bl-young yes, that is a good idea - I'm looking at these pull requests now. 
We could leave this pull request open for now until we do create methods with multi sector levels

catherinebirney:
@bl-young I'd like to pull this branch in because of the fixes to the sector length subsets for household/government codes. The differences in FBS make sense to me - many of the water methods haven't been updated on DC in a while.

bl-young:
agreed, works for me. thanks for fixing that CAP_HAP issue, I think that was from my other PR

PR #173, opened 2022-01-26
Urban rural disaggregation

a-w-beck:
Functions to map flows with FIPS code Location values to Census-data-derived urban population percentages, and then split flows into 'urban' and 'rural' components plus append a secondary Compartment label.

Also migrated a collection of Location- and FIPS-related functions from common.py to location.py.

PR #187, opened 2022-03-08
code updates for v1.2 release

catherinebirney:
- Enables additional, less aggregated, sector levels in final FBS dataset
- When subsetting data based on sector levels, FLOWSA functions generally take the string length of a sector. This method causes issues when the sector is a household or government sector code, where the string length can deviate from the sector level. The revised code subsets dataframes using the sector length crosswalk, which correctly categorizes household/government codes.
- Updates the NAICS crosswalk to include NAICS 7
- Adds an optional FBS parameter activity_to_sector_mapping to point to alternative mapping files.
- Replaces the hard-coded selection for BEA and NEI mapping files.
- Validate FlowBySectors, downloads the latest generated FBS on the remote server and compares the output to one generated locally at the current HEAD. FBS are compared using validation function compare_FBS_results. FBS not available on remote are skipped. For stewi based data sources, requires better access to remote in stewi (https://github.com/USEPA/standardizedinventories/pull/101)
- Functions to map flows with FIPS code Location values to Census-data-derived urban population percentages, and then split flows into 'urban' and 'rural' components plus append a secondary Compartment label.


bl-young:
This includes the additional of urban/rural classification as well, right? (https://github.com/USEPA/flowsa/pull/173)

catherinebirney:
@bl-young Yes thank you, updated the comment
Still todo in this branch:
- change license to MIT
- drop certain FBAs (EQUATES?)
- drop unfinalized FBS (water_state)

PR #189, opened 2022-03-12
add test to read and call urls for FBA methods 

bl-young:
URLs are not tested for methods with API key


bl-young:
@matthewlchambers note the EPA_Equates URL is failing in this action. Is that link still valid?

PR #190, opened 2022-03-12
Refactor process(scc) adjustments by NAICS from stewi files

bl-young:
Adjustments initiated to support reassignment of airplane emissions from NAICS related to airports to air travel are made more general to support other applications.

csv file indicating the `source_naics` (i.e., original facility naics) and the `source_process` are labeled with the `target_naics` to which they are reassigned.

The csv file is identified by name in the stewi FBS parameter `reassign_process_to_sectors` e.g.,
`reassign_process_to_sectors: 'airplane_emissions'`

The file must be stored in the `process_adjustments` folder, but this folder is supported like other inputs outside of the repository (https://github.com/USEPA/flowsa/pull/178)


matthewlchambers (approved):
This seems to work well. My only comment/concern is that as currently implemented, all process adjustments must be in a single file. This will make it harder for someone trying to use the `fbsconfigpath` option to run `flowsa` using external FBS methods, etc.

PR #194, opened 2022-03-23
Allocation fix

catherinebirney:
1. In equally_allocate_parent_to_child_naics(), account for situations where activities are naics-like

2. Modify equal_allocation() to return an error if an activity is being equally allocated to sectors of different lengths

3. BLS_QCEW code
- update the FBA FlowNames to distinguish ownership. This change enables better estimates of suppressed data.
- clean up/remove some old code whose functions are now captured in more automated way within the package (clean_bls_qcew_fba(), replace_missing_2_digit_sector_values(), remove_2_digit_sector_ranges(), bls_clean_allocation_fba_w_sec_sat_table())

4. Modify USDA_CoA_Cropland FBA
- update the FlowName for irrigated, harvested cropland (111) to match the flownames that aggregate to that value
- remove functions that are duplicated or where code is better captured elsewhere(coa_irrigated_cropland_fba_cleanup(), modify_orchard_flowamounts())
- update disaggregate_pastureland() and disaggregate_cropland() so when a parameter is indicated to be dropped (like 1125 (aquaculture) in the water FBS method), that any parent sectors are also dropped, so the df is re-aggregated and outputs more accurate allocation ratios

5. Address error in the Water_national FBS methodology
- For the water FBS, the USGS_NWIS_WU data had to be modified to distinguish between crop and golf irrigation. The function check_golf_and_crop_irrigation_totals() checks if a state distinguishes between crop/golf irrigation and if not, assigns flows to crop irrigation. This function previously did not modify the "Description" in a row, instead, only modified the "Activity" name. By not also modifying the description value, water withdrawals for crop irrigation were double counted in 5 states, as both crop irrigation and total irrigation were included in the dataset. Correcting this error results in a decrease in water withdrawal allocated to crop-related sectors.

6. flowbyfunctions.py changes
- update aggregator() to retain zero flows by default, with option to drop those rows
- new function return_primary_sector_column()
- update the fxn equally_allocate_suppressed_parent_to_child_naics() to allocate data from the nearest unsuppressed parent sector rather than allocating from the nearest parent sector level that contained no suppressed data. This change results in more accurate data allocation, where child naics do not sum to be greater than parent naics (unless the way data was rounded in the orignal df results in some minor differences). Also add validation checks within the function to compare original flow amounts at each sector level to the flow amounts in the revised df. A second validation check compares summed child sectors to parent sector values.
- new fxn assign_columns_of_sector_levels() to automate determining sector levels

7. Update functions throughout to use subset_df_by_sector_lengths() to subset a df by subset lengths, rather than taking the string length of a sector column. This change prevents errors caused by the nontraditional government and household sector codes where the string length does not always match the sector length assignment.

8. Recreated the BLS_QCEW activity set file using 2010-2018 files. This update added some sectors that were previously dropped.

9. New state level employment FBS for 2002, 2012-2017

10. validation.py
- new function compare_child_to_parent_sectors_flowamounts() to check the summed values of child naics vs parent flows


bl-young:
@catherinebirney is there any error handling that could be added so that its clearer what's going on currently with the actions? I assume its because the allocation flow names do not exist in the dataset (which seems like an error that could appear in the future as well)

catherinebirney:
@bl-young This branch is ready for review (as long as the tests comparing FBS to those on Data Commons is successful)

bl-young:
> @bl-young This branch is ready for review (as long as the tests comparing FBS to those on Data Commons is successful)

It's getting hung up on the test FBA config for some reason. I will cancel and re-start

catherinebirney:
@bl-young thanks for taking care of those errors. I've addressed additional issues, so this PR is ready for review (again, assuming the FBS builds do not fail)

bl-young:
@catherinebirney looks like all the FBS ran. Scrolling through, I still see some that show 100% difference from the FBA, primarily in the water methods. Is that to be expected?

catherinebirney:
> @catherinebirney looks like all the FBS ran. Scrolling through, I still see some that show 100% difference from the FBA, primarily in the water methods. Is that to be expected?

That is odd because when I run the code locally, all of those FBS flows are equal to the FBA values. And the data loss is occurs for an activity set that is direct allocation, so it isn't that the data changed. I will investigate...

catherinebirney:
@bl-young pandas 1.4.2 interpreted df dtypes differently than previous versions of pandas. The latest commit should fix the issues where all values for direct allocation in water were dropped.

bl-young:
> @bl-young pandas 1.4.2 interpreted df dtypes differently than previous versions of pandas. The latest commit should fix the issues where all values for direct allocation in water were dropped.

ahh great! where's the fingers crossed emoji?
Do you have a link to documentation of that changes in pandas (I'm just curious to learn more)

catherinebirney:
I didn't find documentation changes, in fact, based on documentation, nothing changed - I narrowed down the issue to a single pd.merge function that worked with pandas 1.3.5 but not 1.4.2.

catherinebirney:
@bl-young @WesIngwersen I spoke with someone from USGS who says the mineral team is having issues with their website and they are working on it. So the failing FBA links are not on our end - none of the links have changed (yet)

bl-young:
> Do you know why there aren't comparisons created for Employment_state FBS? There are parquets on data commons and there are data differences.

I looked at one and it said no differences. So it won't create the csv.

catherinebirney:
> I looked at one and it said no differences. So it won't create the csv.

Ha - sorry! That's right, I uploaded versions on Wednesday for Mo, so there shouldn't be changes since then.

bl-young (approved):

PR #196, opened 2022-03-30
Adjustments to schema.py and sectormapping.py

matthewlchambers:
- Made some adjustments to improve readability of `sector_mapping.add_sectors_to_flow_by_activity()`.
- Wrote a new function `common.get_catalog_info()` to replace the previous `common.get_true_source_name()`, which was only ever used as a precursor to loading the source infor from `source_catalog.yaml`.
- Refactored `schema.py` so that the entries in `activity_fields` are dictionaries instead of lists of single-entry dictionaries.


matthewlchambers:
Abandoned this line of work.

PR #197, opened 2022-03-30
NAICS dictionary for Summary models

bl-young:
Add a NAICS dictionary relevant for Summary level models which can be accessed using:
`inherits_from: BEA_Summary_target` in any FBS


bl-young:
@MoLi7, @WesIngwersen, I believe @catherinebirney and I confirmed that these are the only sub-3 digit NAICS needed for summary models. With the exception of handling 531*, 532*, 533*, and 92*, which are mixed BEA sectors at all levels and will need to be handled in useeior no matter the target (and therefore _I believe_ 3 digit should still be fine)

catherinebirney (approved):

PR #199, opened 2022-04-06
enable multiple `inherits_from` in fbs method

bl-young:
uses [dpath](https://pypi.org/project/dpath/) for better accessing and manipulating of dictionaries in order to support `inherits_from` at different nodes in FBS method.

For [example](https://github.com/WesIngwersen/GHGbyindustrydocs/commit/b33a3dec3209096ac13d6c66885cebae2d3ad901):
```
inherits_from: BEA_summary_target
target_geoscale: national
source_names:
  inherits_from: GHG_national_2016_m1
```

Using inherits from in an FBS method will update the method from the target identified at that specific node. In the example above, everything within `source_names` from `GHG_national_2016_m1` is added to this method. No further adjustments of this node are possible.

This PR also ensures that inherits_from can use target methods within local directory (first) before looking for the target method in the repo. In the example above prevents the need to have a local version of `BEA_summary_target` when generating FBS outside the repo.

Existing use of `inherits_from` in FBA methods are not impacted.

### DIFF
--- a/flowsa/common.py
+++ b/flowsa/common.py
@@ -9,6 +9,7 @@ import os
 import yaml
 import pandas as pd
 import numpy as np
+import dpath.util
 from dotenv import load_dotenv
 from esupy.processed_data_mgmt import create_paths_if_missing
 from flowsa.schema import flow_by_activity_fields, flow_by_sector_fields, \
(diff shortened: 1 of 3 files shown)


bl-young:
fyi @WesIngwersen I tested the [example summary method](https://github.com/WesIngwersen/GHGbyindustrydocs/commit/b33a3dec3209096ac13d6c66885cebae2d3ad901) above and the results at a glance seem to match the totals of the detailed model

catherinebirney (approved):

matthewlchambers:
Does this change the order of precedence? That is, does the parent now override the child, rather than the other way around?

bl-young:
> Does this change the order of precedence? That is, does the parent now override the child, rather than the other way around?

I think it will, which may not be desired. There may be a way to rearrange that. It certainly would be nice if we could keep any data at that node from the child and not have all the parent override it. Something like the following (which I don't think is correct FBS method, but would allow for revising the year while keeping everything else from the parent)
```
inherits_from: BEA_summary_target
target_geoscale: national
source_names:
  year: 2018
  inherits_from: GHG_national_2016_m1
```
I'm open to other ideas to implement this.

matthewlchambers:
I'll think about it. I've been looking into custom `yaml` tags and constructors anyway, so maybe something along those lines will work. Just to make sure I understand correctly, the goal is for us to be able to inherit a particular node from another file, at any level of the tree?

bl-young:
> Just to make sure I understand correctly, the goal is for us to be able to inherit a particular node from another file, at any level of the tree?

Correct. And to ensure that we can inherit from more than one file.

matthewlchambers:
See my implementation in #200. It would require some refactoring of existing `yaml` files, but I think it incorporates the best features of both my previous implementation and this one, as well as making further changes in the future easier.

matthewlchambers (changes requested):
See my proposed alternate implementation in https://github.com/USEPA/flowsa/pull/200. It would require some refactoring of existing `yaml` files, but I think it incorporates the best features of both my previous implementation and this one, as well as making further changes in the future easier.

bl-young:
closing in favor of #200

PR #200, opened 2022-04-11
Added custom yaml loader

matthewlchambers:
Added a `flowsa_yaml` module including a custom `yaml` loader called `FlowsaLoader`, and a function `flowsa_yaml.load(stream: IO, external_config_path: str = None)`. The `load` function can be used similarly to the existing `yaml.safe_load()` function, as demonstrated below. The custom loader introduces 2 features:
1. An `!include:<file>:<key>: . . . :<key>` tag is made available for use in our `yaml` method files. A key may be labeled with this tag to include a specific node from a given file. The precise way in which the inclusion occurs depends on the type of node the `!include:` tag is used on: if the tagged node is a scalar node, the included node will be added as the value for the tagged node; if the tagged node is a sequence/list node, the included node (which must also be a sequence/list) will be appended to the beginning of the tagged node's list; and for a mapping/dict node, the included node (which must also be a mapping/dict) will be updated with the key/value pairs from the tagged node (so any matching keys will have the value from the tagged node). If the user attempts to `!include:` a non-mapping node on a mapping node, or a non-sequence node on a sequence node, a `TypeError` will be raised.
2. An `external_config_path` paramter may be supplied to `flowsa_yaml.load()` or the `FlowsaLoader` class. At present, only one may be supplied, and its only use is to specify a folder in addition to flowsa's internal `flowbyactivitymethods/` and `flowbysectormethods/` folders where the `!include:` tag can find a file.

Additionally, this will give us an easy way to implement any other custom tags or other customized handling of `yamls` that may be useful.

Example (supposing we wanted to combine the GHGI sources with NEI onroad data for some reason, but use a custom activity set file for the latter source):

```
source_names: !include:GHG_national_2016_m1.yaml:source_names
  NEI_Onroad: !include:CAP_HAP_national_2017.yaml:source_names:EPA_NEI_Onroad
    activity_set_file: path/to/custom/activity_set_file.csv 
```

This does require refactoring so that `yaml.safe_load()` is replaced in script files with `flowsa_yaml.load()`.

Known issues:
1. I believe `yaml` tags cannot contain spaces, so this implementation won't work with filenames or `yaml` keys that contain spaces. Fortunately, I believe we have generally ensured that filenames and `yaml` keys don't have spaces, but it is something to be aware of.
2. This implementation may not be the best way to handle an external configuration path. In particular, it might be good to have an `!external_config_path:` custom tag to specify such external configuration paths. I believe this would be straightforward to implement.

### DIFF
--- a/flowsa/flowsa_yaml.py
+++ b/flowsa/flowsa_yaml.py
@@ -35,13 +35,19 @@ def include(loader: yaml.Loader, suffix: str, node: yaml.Node) -> dict:
     while keys:
         branch = branch[keys.pop(0)]
 
-    if isinstance(node, yaml.MappingNode) and isinstance(branch, dict):
-        context = loader.construct_mapping(node)
-        branch.update(context)
+    if isinstance(node, yaml.MappingNode):
+        if isinstance(branch, dict):
+            context = loader.construct_mapping(node)
+            branch.update(context)
+        else:
+            raise TypeError(f'{suffix} is not a mapping/dict')
 
-    elif isinstance(node, yaml.SequenceNode) and isinstance(branch, list):
-        context = loader.construct_sequence(node)
-        branch.extend(context)
+    elif isinstance(node, yaml.SequenceNode):
+        if isinstance(branch, list):
+            context = loader.construct_sequence(node)
+            branch.extend(context)
+        else:
+            raise TypeError(f'{suffix} is not a sequence/list')
 
     return branch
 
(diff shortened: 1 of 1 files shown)


catherinebirney:
@bl-young yes, I was thinking the same

bl-young:
@matthewlchambers I will pull this in, feel free to make any more changes directly and I will try to do some testing with some of our methods

matthewlchambers:
Sounds good!

PR #206, opened 2022-04-13
add action to generate all FBS and store as artifact on PR to master

bl-young:
Downloads FBAs rather than generate them locally


bl-young:
I could not pull this directly into master as I needed one update to `seeAvailableFlowByModels` which is already in this branch.

We can test fully once 1.2 is merged in to master

catherinebirney (approved):

PR #207, opened 2022-04-18
Custom yaml loader

bl-young:
Enables custom yaml loader to include keys:
`!include:<file>:<key>: . . . :<key>`: integrates nodes from passed file
`!external_config`: adds additional target paths to search for included files

See https://github.com/USEPA/flowsa/pull/200 and https://github.com/USEPA/flowsa/pull/204 for additional details.


bl-young:
@matthewlchambers , @catherinebirney I think this is probably ok to pull in now?

catherinebirney (approved):
Yes, good to pull in

matthewlchambers (approved):

PR #209, opened 2022-04-20
pass external_path supplied as arg recursively

matthewlchambers:
Tweaked `flowsa_yaml` so that when it is passed an external configuration path by `common.py` (or anything else), it passes that along to the loaders for any inherited `yaml`s. No changes in how the function is called are required (e.g. in `common.py`)


bl-young (approved):
makes sense to me, thanks. i'll revert on my branch and will pull this in instead

PR #210, opened 2022-04-21
Tweak custom yaml loader

matthewlchambers:
Implements a new custom tag `from_index:<file> <key in .csv index>`, which allows us to get lists of activity set names from a `.csv` file while loading the method `yaml`. This would allow us to remove quite a bit of extraneous code from `flowbysector.py`, and parallel our use of the `!include:` tag.

The `from_index:` tag is currently implemented similarly but not identically to the `!include:` tag. Like the `!include:` tag, a file name is included as part of the tag after the colon. Unlike the `!include:` tag, the key to look up is given as the value of the node in the `yaml` file. This is to allow for spaces in the key, which will in turn allow us to create index files that use more descriptive keys than `activity_set_1`, and which would not need to be updated if we changed the order or designation of activity sets for some reason.

An example is given below:
```
activity_set_1:
  names: !from_index:NEI_Nonpoint_2017_asets.csv activity_set_1
```

### DIFF
--- a/flowsa/flowsa_yaml.py
+++ b/flowsa/flowsa_yaml.py
@@ -2,6 +2,7 @@ from typing import IO, Any
 import yaml
 import flowsa.settings
 from os import path
+import csv
 
 
 class FlowsaLoader(yaml.SafeLoader):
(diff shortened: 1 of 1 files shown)


bl-young:
closing this in favor of #213 which brings these changes in with application

PR #213, opened 2022-04-25
Refactor NEI

matthewlchambers:
This refactoring implements #210. Additionally it:
* Refactors all FBS method `yaml`s that use an external `csv` files as an index of names to incorporate #210
* Refactors the above `yamls`s, along with their associated `csv`s, to use descriptive activity set names, rather that just `activity_set_1`, etc.
* Refactors `CAP_HAP_national_2017.yaml` to split activity set 3 into separate activity sets, such that the subsetting columns from the `NEI_Nonpoint_2017_asets.csv` file are no longer necessary.
* Removes the subsetting columns from `NEI_Nonpoint_2017_asets.csv`
* Refactors `flowbysector.py`, `fbs_allocation.py`, and `sectormapping.py` to remove the code that used the subsetting solumns previously mentioned.

These changes, particularly the second, did result in changes to the generated FBS. However, according to @bl-young, they represent the original intent behind the code that used the subsetting columns, and reduce the size of the discrepancy between the FBS total and the totals from the source FBAs.


flowsa/methods/flowbysectoractivitysets/NEI_Nonpoint_2017_asets.csv
```diff
@@ -1,594 +1,594 @@
...
+cropland,2801000008,Agriculture - Crops
+cropland,2801500000,Agriculture - Field Burning
+cropland,2801500151,Agriculture - Field Burning
+cropland,2801500152,Agriculture - Field Burning
+cropland,2801500161,Agriculture - Field Burning
+cropland,2801500263,Agriculture - Field Burning
+cropland,2801500264,Agriculture - Field Burning
+cropland,2801500300,Agriculture - Field Burning
+cropland,2801500360,Agriculture - Field Burning
+industry_combustion_a,2102001000,Fuel Combustion - Industry
```
bl-young, line 56: for consistency with the other updates, maybe we should just go ahead and label these with the fuel type like `industry_combustion_coal`

flowsa/methods/flowbysectormethods/Employment_national_2017.yaml
```diff
@@ -16,7 +16,8 @@ source_names:
     clean_fba_w_sec_df_fxn: bls_clean_allocation_fba_w_sec_sat_table
     activity_set_file: 'BLS_QCEW_asets.csv'
```
bl-young, line 17: this one can be dropped as well

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
```diff
@@ -28,6 +22,28 @@ _allocation_types:
     helper_from_scale: national
     clean_helper_fba: clean_bls_qcew_fba
     clean_helper_fba_wsec: bls_clean_allocation_fba_w_sec
+  _use_allocation: &use_allocation
```
bl-young, line 25: can you add a blank line above this use_allocation for visual effect?

also, I like how `clean_parameter` is shown but commented out, the same as `allocation_flow` above for EIA_MECS. Maybe you could add the `names` parameter as well to both, as a comment, as a flag that it needs to be included for an individual `activity_set`

flowsa/methods/flowbysectoractivitysets/NEI_Nonpoint_2017_asets.csv
matthewlchambers: That's actually what I started doing, but some of the fuel types, like "Hydrocarbon gas liquids, excluding natural gasoline" are pretty long, and using only a portion of the name, like `hydrocarbon_liquids`, implies something that isn't true. Thoughts?

matthewlchambers:
> does this mean that the FBS parameter `activity_set_file` would no longer be needed at all?
> 
> I'll flag that @catherinebirney tends to [use scripts](https://github.com/USEPA/flowsa/tree/master/scripts/FlowBySector_Activity_Sets) to generate the activity sets, so those may need to be updated but should be straight forward.

That's correct re: the `activity_set_file` parameter. And I was wondering about those scripts, but haven't taken a look at them yet.

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
matthewlchambers: Good ideas!

flowsa/methods/flowbysectoractivitysets/NEI_Nonpoint_2017_asets.csv
matthewlchambers: Nevermind, just pushed an update with better names.

matthewlchambers: Good point!

matthewlchambers:
I want to clarify that, as indicated in #210, the name of the activity set in the `yaml` file does _not_ need to match how it is labeled in the `csv` file, even though in this refactor I've had them match in this way.

bl-young (approved):

bl-young:
@catherinebirney are we ready to pull this in? Did you want to update your scripts and/or the [FBS readme](https://github.com/USEPA/flowsa/tree/refactor-nei/flowsa/methods/flowbysectormethods#readme) before doing so?

I am considering pulling this in to `state_ghg` so I can use these updates as I work on that method

matthewlchambers:
> @catherinebirney are we ready to pull this in? Did you want to update your scripts and/or the [FBS readme](https://github.com/USEPA/flowsa/tree/refactor-nei/flowsa/methods/flowbysectormethods#readme) before doing so?
> 
> I am considering pulling this in to `state_ghg` so I can use these updates as I work on that method

Just updated the scripts and readme, so I believe this pull request can be considered complete.

catherinebirney:
@matthewlchambers Thanks for making those changes
@bl-young yes, ready to pull this in

PR #214, opened 2022-05-03
Tweak custom yaml loader

matthewlchambers:
Adds another new tag to the custom `yaml` loader, to remove the need for the `dynamically_import_fxn` function. Instead, nodes in the `yaml` file that provide the name of a function from `<data_source_module>.py` to be used in processing that specific data source may be tagged with `!script_function:<data_source_module>`. The function name is then given as the node value. An example is given below:

```
FBS_datapull_fxn: !script_function:stewiFBS stewicombo_to_sector
```

In this example, the loader will then import `stewiFBS` and attach the named function at the correct spot in the config dictionary, so that the function may then be called as `source_config['FBS_datapull_fxn'](yaml_load, method, fbsconfigpath)`.

I believe that this change will make both the configuration `yaml` files and the codebase easier to read and follow.

As currently implemented, only modules installed with `flowsa` in the data source scripts folder can be imported this way. I'm waiting to implement searching the external configuration path until I can do some more research on security implications of doing so.


bl-young:
I believe the first two commits are now already pulled in so the only change here is [1688139](https://github.com/USEPA/flowsa/pull/214/commits/16781398fd76bd75ab3dead8bc3f79be539c6c53).

So the upshot here is that this would also enable using functions agnostic to the data source (since currently `dynamically_import_fxn` uses the module for the data source itself). In addition to the readability improvements you suggest.

There is at least one instance of where that has to be done manually:
https://github.com/USEPA/flowsa/blob/a83eb1e8788b29a29ddea93cd5d4c82460f4b394/flowsa/methods/flowbysectormethods/CNHW_national_2014.yaml#L7-L8

There are a number of locations where `dynamically_import_fxn` is used throughout so each would need to be updated to this new system, along with most of the FBS. If I'm interpreting correctly, using `!script_function:` would not work as currently structured (since it would break when it hits `dynamically_import_fxn`

matthewlchambers:
> I believe the first two commits are now already pulled in so the only change here is [1688139](https://github.com/USEPA/flowsa/pull/214/commits/16781398fd76bd75ab3dead8bc3f79be539c6c53).
> 
> So the upshot here is that this would also enable using functions agnostic to the data source (since currently `dynamically_import_fxn` uses the module for the data source itself). In addition to the readability improvements you suggest.
> 
> There is at least one instance of where that has to be done manually:
> 
> https://github.com/USEPA/flowsa/blob/a83eb1e8788b29a29ddea93cd5d4c82460f4b394/flowsa/methods/flowbysectormethods/CNHW_national_2014.yaml#L7-L8
> 
> There are a number of locations where `dynamically_import_fxn` is used throughout so each would need to be updated to this new system, along with most of the FBS. If I'm interpreting correctly, using `!script_function:` would not work as currently structured (since it would break when it hits `dynamically_import_fxn`

I concur with everything you said here.

catherinebirney:
Great suggestion @matthewlchambers, I'm on board. Although I like that this change will improve the clarity of where the function is imported from, I agree with Ben that the greatest benefit is being able to call on any function.

bl-young:
I assume we should update all of the FBS and the `dynamically_import_fxn` within this branch before pulling it in?

matthewlchambers:
> I assume we should update all of the FBS and the `dynamically_import_fxn` within this branch before pulling it in?

I agree as well. I'll change this to a draft PR until that is completed.

matthewlchambers:
Finished the refactor, so I believe this is ready for review/merging, @bl-young @catherinebirney . Thanks!

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
```diff
@@ -25,8 +25,8 @@ source_names:
           - "AREA HARVESTED"
         allocation_compartment: None
         allocation_from_scale: state
-        clean_allocation_fba: coa_irrigated_cropland_fba_cleanup
-        clean_allocation_fba_w_sec: disaggregate_coa_cropland_to_6_digit_naics
+        clean_allocation_fba: !script_function:EPA_NEI coa_irrigated_cropland_fba_cleanup
+        clean_allocation_fba_w_sec: !script_function:EPA_NEI disaggregate_coa_cropland_to_6_digit_naics
       activity_set_3: #Fuel Combustion - Industry #only allocating to 3digits
```
bl-young, line 30: for these clean functions, the script source is the allocation source, e.g.:
`!script_function:USDA_CoA_Cropland coa_irrigated_cropland_fba_cleanup`

bl-young:
@matthewlchambers I'll review the FBS per my comment above and push a few changes

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
matthewlchambers: > @matthewlchambers I'll review the FBS per my comment above and push a few changes

Ah, of course. Thanks for taking care of that!

bl-young:
@matthewlchambers https://github.com/USEPA/flowsa/pull/214/commits/cae43066a27bfad694a42cd839cc589c9b9156ce should ensure that all methods can load correctly

I will bring in and merge develop into this branch in order to bring in #213

bl-young (approved):

catherinebirney (approved):
I like this change a lot, thanks @bl-young @matthewlchambers!

PR #218, opened 2022-05-12
Use new `yaml` tag for `EPA_NEI.drop_GHGs`

matthewlchambers:
Updated `EPA_NEI.py`, `stewiFBS.py`, and `CAP_HAP_national_2017.yaml` to make use of the new `!script_function:` tag and use `drop_GHGs` as a `clean_fbs_df_fxn`, since that's commonly used in flowsa yamls.

I removed some of the extra code in `stewicombo_to_sector` that called the `drop_GHGs` function.

I made it so the `drop_GHGs` function can operate on columns named either `FlowName` or `Flowable`.


catherinebirney (approved):

flowsa/data_source_scripts/EPA_NEI.py
```diff
@@ -208,26 +208,22 @@ def remove_duplicate_NEI_flows(df):
...
         'Carbon Dioxide',
         'Methane',
         'Nitrous Oxide',
         'Sulfur Hexafluoride',
     ]
-
-    df = df.loc[~df['FlowName'].isin(flowlist)]
-
-    return df
+    flow_var = 'Flowable' if 'Flowable' in df.columns else 'FlowName'
```
bl-young, line 225: I'll need to confirm this is fine - we use Flowable after mapping, and FlowName before IIRC. There are syntax and capitalization issues that make this distinction important.

flowsa/data_source_scripts/stewiFBS.py
```diff
@@ -43,12 +43,8 @@ def stewicombo_to_sector(yaml_load, method, fbsconfigpath=None):
     :param fbsconfigpath, str, optional path to an FBS method outside flowsa repo
     :return: df, FBS format
     """
-
     import stewicombo
-    from flowsa.data_source_scripts.EPA_NEI import drop_GHGs
 
-    # determine if fxns specified in FBS method yaml
-    functions = yaml_load.get('functions', [])
```
bl-young, line 51: If I recall we had added this "functions" to allow for some customized additional steps for some stewi based FBS, that we ultimately didn't use.

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
```diff
@@ -196,9 +196,8 @@ source_names:
     FBS_datapull_fxn: stewicombo_to_sector
     inventory_dict: {"NEI":"2017", "TRI":"2017"}
     local_inventory_name: 'NEI_TRI_air_2017'
+    clean_fbs_df_fxn: !script_function:EPA_NEI drop_GHGs
```
bl-young, line 199: `clean_fbs_df_fxn` didn't exist at the time of this method. I believe this change should be fine, but I'd like to confirm that keeping the GHGs until later (i.e. after `prepare_stewi_fbs` doesn't cause any issues).

bl-young (commented):
These changes make sense in theory to me. Though I think I'd prefer to wait until after the v1.2 release/ `allocation_fix` branch, as this would be a good opportunity to confirm that these changes lead to no changes in FBS (which I believe is the expected outcome). The action will help us determine this.

flowsa/data_source_scripts/EPA_NEI.py
matthewlchambers: Yes, that's correct. The goal of line 225 was to make the function suitable for use either before or after mapping.

flowsa/data_source_scripts/stewiFBS.py
matthewlchambers: Makes sense. My thought is that we can still pull this code out (it's not used for anything in `flowsa` besides `drop_GHGs`), and then if customized additional steps are desired in the future we have some new tools that can be used to define those additional steps. (For instance, to arrive at very similar functionality to the old code, we could modify the custom `yaml` loader to allow a list of functions to be passed after the `!script_function:` tag.)

flowsa/methods/flowbysectormethods/CAP_HAP_national_2017.yaml
matthewlchambers: Agreed.

flowsa/data_source_scripts/EPA_NEI.py
bl-young: Right - but the flow names in the list above are specific to the NEI pre-mapping, as the names update after mapping.

matthewlchambers: Ah, makes sense.

matthewlchambers: Ah, makes sense.

bl-young:
@matthewlchambers the current comparison to the FBS on data commons is not working quite right because the stewi-based files up there are not the right version. I am working to fix that.

However I was able to compare the results of this PR on the CAP_HAP method and the result is that this _leaves_ the GHGS in the dataset due to the naming issue I mentioned above. [CAP_HAP_diff.csv](https://github.com/USEPA/flowsa/files/8790145/CAP_HAP_diff.csv) 
All else is equal.

So it seems like the solution is either to expand the flow list above to account for the flow names before AND after mapping. Or we move that list to the method yaml instead (perhaps as a more generic 'drop_flows' function).

matthewlchambers:
> @matthewlchambers the current comparison to the FBS on data commons is not working quite right because the stewi-based files up there are not the right version. I am working to fix that.
> 
> However I was able to compare the results of this PR on the CAP_HAP method and the result is that this _leaves_ the GHGS in the dataset due to the naming issue I mentioned above. [CAP_HAP_diff.csv](https://github.com/USEPA/flowsa/files/8790145/CAP_HAP_diff.csv) All else is equal.
> 
> So it seems like the solution is either to expand the flow list above to account for the flow names before AND after mapping. Or we move that list to the method yaml instead (perhaps as a more generic 'drop_flows' function).

Thanks for doing that comparison. I think for now the best approach would be to just add the names before mapping to the `drop_GHGs` function. I can go ahead and do that, though it may be a couple of days before I can, as I'm heading out of town for a conference.

bl-young:
> Thanks for doing that comparison. I think for now the best approach would be to just add the names before mapping to the `drop_GHGs` function. I can go ahead and do that, though it may be a couple of days before I can, as I'm heading out of town for a conference.

Works for me, I should be able to do that no problem and then can re-run to compare.

bl-young (approved):
https://github.com/USEPA/flowsa/runs/6696679230?check_suite_focus=true#step:7:3101
> ***No differences found in CAP_HAP_national_2017***

bl-young:
ready to merge as soon as we get write access again

Issue #223, opened 2022-05-17
Census_CBP missing data


bl-young:
Census_CBP is only generating `Number of establishments` data

https://github.com/USEPA/flowsa/blob/715d8e9eb487b2250539bce4c5545baeb9d72758/flowsa/data_source_scripts/Census_CBP.py#L145-L147

`EMP` and `PAYANN` are not showing up in the data call

PR #224, opened 2022-05-17
add employment and annual pay to Census_CBP

catherinebirney:


catherinebirney:
@bl-young At some point employment and annual pay were commented out. 

Solves issue [223](https://github.com/USEPA/flowsa/issues/223)

bl-young:
great thanks! If it runs for you successfully then I'm fine to bring it in.

PR #225, opened 2022-05-18
allocate regional mecs data to states

bl-young:
Adds a `clean_fba_w_sec` function when using EIA MECS as allocation source so that the data from the four census regions is apportioned to states proportionally to employment by sector.

Accessed via `eia_mecs_energy_clean_allocation_fba_w_sec_to_state()`


flowsa/flowbyfunctions.py
```diff
@@ -527,7 +527,7 @@ def subset_df_by_geoscale(df, activity_from_scale, activity_to_scale):
         cols_to_keep = flow_by_activity_fields
 
     # method of subset dependent on LocationSystem
-    if df['LocationSystem'].str.contains('FIPS').any():
+    if df['LocationSystem'].str.contains('FIPS').all():
```
bl-young, line 530: @catherinebirney I am curious as to your thoughts on the two changes in this module.

EIA_MECS FBA has both FIPS and Census_Region as a location system. The U.S. data is assigned as `00000`

Are there other cases where an FBA will have multiple location systems?

I'm realizing now what I have here is going to cause problems with the national MECS data as allocation source, will need to rethink this...

PR #226, opened 2022-05-25
add `USEEIO_summary_target` method with 221 disaggregated

bl-young:


bl-young:
Attached file shows successful update to 221* from 221 for the water method.
[Water_state_2015_m1_diff.csv](https://github.com/USEPA/flowsa/files/8790155/Water_state_2015_m1_diff.csv)

Diff file for CNHW_state suffers from differences in use of flow mapping, including unit conversion so needs to be reviewed further.

bl-young:
This PR can be re-assessed after CNHW is updated on data commons following #232

bl-young:
When compared with the current HEAD (not on datacommons as of this PR), diff is as expected showing disaggregation of 221 into 2211 and 2213. 
[CNHW_state_2014_dif.csv](https://github.com/USEPA/flowsa/files/8837809/CNHW_state_2014_dif.csv)
Note that CA has no employment for 2212 so due to the method for national and state FBS no jobs are assigned to that sector in any state

PR #230, opened 2022-06-01
Update BEA data with latest data

bl-young:
- Updates 2012 Make and Use tables to use latest BEA release in Fall of 2021 (see https://github.com/USEPA/useeior/pull/221)
- Updates industry gross output data from same release, and expands to include 2019 and 2020 data years
- deprecated `BEA_2012_Detail_Use_Industry_Transactions.csv` removed to avoid confusion


bl-young:
cc @catherinebirney tagging you so you are aware

MoLi7 (approved):

bl-young:
Due to the expected changes in FBS, wait until #231 is merged

bl-young:
@MoLi7 I noticed that the new BEA data for Gross Output only has 404 sectors instead of 405, perhaps due to that Drugs and druggists' sector issue you recently fixed (after I pulled in the data). I'll propose that we wait until useeior has been released before we confirm those updates here and pull this in.

MoLi7:
Yes, the issue Ben raised has been fixed in useeior v1.1.0.

Issue #216, opened 2022-05-10
Custom exceptions


bl-young:
It might be nice to start using more custom exceptions for better code traceback and error handling.

This would predominantly apply in places where we use `log.error` for code breaking issues.

matthewlchambers:
In some cases, we may not even need custom exceptions, just using `try: . . . except <Exception>:` blocks, with `log.exception` in the `except` clause. According to [this documentation](https://docs.python.org/3/library/logging.html#logging.exception), that should include the traceback along with our desired logging error message.

matthewlchambers:
Made a couple of comments re: naming of the custom exceptions in cd02d0187766898abcaf36b3e636e19bba56c9b1 . I think we want to be careful that we're not adding custom exceptions when writing better log messages would be just as helpful.

bl-young:
> Made a couple of comments re: naming of the custom exceptions in [cd02d01](https://github.com/USEPA/flowsa/commit/cd02d0187766898abcaf36b3e636e19bba56c9b1) . I think we want to be careful that we're not adding custom exceptions when writing better log messages would be just as helpful.

great thanks. feel free to point out any times where you think a better log exception would be better - I'm not sure I fully comprehend the difference. I'm trying to start with errors i've encountered and/or found in the gh actions that tend to hide the true issue

PR #233, opened 2022-06-04
Update error handling

bl-young:
Including through the use of custom exceptions in order to better track the source of issues. Common errors include:

- missing API key
- method unavailable (e.g. can't be found or incorrect name)
- errors in calling script functions
- errors in FBS method construction (e.g. allocation data subset incorrectly)

Implements #216


bl-young:
diff results are as expected

Issue #238, opened 2022-07-08
in geo.filtered_fips: TypeError: unhashable type: 'Series'


bl-young:
Obtaining the national FIPS code in `geo.filtered_fips` returns a TypeError

https://github.com/USEPA/flowsa/blob/a45e737d02a77d10cd358913cce5f4d13f7bef41/flowsa/geo.py#L89-L90

See for example this example code, fails on the third attempt.
```
import flowsa.geo as geo

geo.get_all_fips(2015).query('State != State') # option 1 see https://stackoverflow.com/a/46021212
geo.get_all_fips(2015).query('State.isnull()', engine='python') # option 2
geo.get_all_fips(2015).query('State.isnull()') # current method fails
```

I'm using pandas 1.3.3 and its possible this bug may be fixed in [pandas 1.4](https://github.com/pandas-dev/pandas/issues/22435). See [this comment](https://github.com/pandas-dev/pandas/issues/22435#issuecomment-916479899).

We either need to require pandas 1.4 or implment one of the options above

matthewlchambers:
If we want to maintain pandas 1.3.3 compatibility, I prefer option 2 as a fix. However, I think requiring pandas 1.4 is probably reasonable (in line with shifting the python requirement to 3.8), especially since my understanding is that 1.4 fixed quite a lot of bugs.

matthewlchambers:
Having done a little more research on this issue (to understand what other issues might be caused by this bug), I feel more strongly than before that it would be better to require pandas 1.4. For pandas < 1.4, this bug causes (or may cause) `.query()` to fail if the query string uses pandas `Series` methods. Option 1 is a fix that is very specific to the situation of checking for nulls, and option 2 (`engine='python'`) has performance implications which may be annoying with larger datasets.

On the whole, particularly given that this issue arises because of a bug which got fixed in pandas 1.4, it would be better to require 1.4 than to make potentially awkward fixes to accommodate the bug.

bl-young:
Noting that pandas 1.4 requires python >= 3.8, so relates to #237

bl-young:
Ok in that case https://github.com/USEPA/flowsa/commit/c6d261e982e4759e4a622d3080dcfa512345ea25 can be reverted and we can make a plan to update requirements (with hopefully few other issues)

PR #244, opened 2022-07-11
update pandas requirement to 1.4.0 and Python to >=3.8

catherinebirney:
- update pandas requirement from 1.3.2 to 1.4.0
- drop support for testing FLOWSA package on Python 3.7
-  addresses issues discussed in Discussion [237](https://github.com/USEPA/flowsa/discussions/237) and Issue [328](https://github.com/USEPA/flowsa/issues/238)

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        py-version: ['3.7', '3.8', '3.9', '3.10']
+        py-version: ['3.8', '3.9', '3.10']
 
     steps:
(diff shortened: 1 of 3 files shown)


catherinebirney:
@matthewlchambers @bl-young

bl-young:
I don't know if this is the right way to go or not, but would it be worth specifying the minimum python version in setup.py?
https://stackoverflow.com/a/48777286, or anywhere else for that matter

catherinebirney:
@bl-young I was thinking the same. We do have the 3.7 requirement listed in the [wiki](https://github.com/USEPA/flowsa/wiki/Install-&-Run), but might not be an obvious place to work

catherinebirney:
See PR #251 for explanation on FBS differences

PR #246, opened 2022-07-11
Adds display_tables() method to FBS

matthewlchambers:
This PR adds to `FlowBySector` a `display_tables()` method which generates and saves (as a multi-sheet Excel file) nice display tables, which may be customized (if desired) in the FBS method_yaml, like this:
```
display_tables:
  <Table 1 title>:
    selection_fields:
      # exactly as used elsewhere
    industry_spec:
      # formatted as elsewhere, but can be used to produce a table that is more aggregated than the underlying FBS
    replace_dict:
      <column>:
        <to_replace>: <new_value>
        <to_replace>: <new_value>
        # etc.
```

### DIFF
--- a/flowsa/data_source_scripts/EPA_GHGI.py
+++ b/flowsa/data_source_scripts/EPA_GHGI.py
@@ -680,7 +680,7 @@ def get_manufacturing_energy_ratios(year):
     return pct_dict
 
 
-def allocate_industrial_combustion(fba, source_dict, **_):
+def allocate_industrial_combustion(fba, *_, **__):
     """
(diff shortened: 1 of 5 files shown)


bl-young:
@matthewlchambers I saw in one of your recent commits you were working on a summary table by BEA sector. I wanted to alert you to [this function](https://github.com/USEPA/flowsa/blob/97fdab9250d205bfd8d9341e3b62cc464bc854c7/flowsa/sectormapping.py#L445-L500) which can be used to map a df from NAICS to BEA. It may or may not be useful to you.

PR #247, opened 2022-07-12
Geoscale

catherinebirney:
update allocation methodology so if the allocation dataset is at a more aggregated geoscale than the primary FBA, that the primary df is not aggregated to match the allocation geoscale. Instead the more aggregated allocation dataset merges with the primary FBA on all related less aggregated geoscales

Potentially impacts state level models

Includes changes to github actions to address memory issues - gh actions now tests all FBS methods without stopping due to memory issues


bl-young:
so the state water method isn't affected here?

matthewlchambers:
FYI, the approach (to cases where the attribution dataset is more aggregated than the source dataset) in the `seea-recursive` branch is very similar to what @catherinebirney has here.

bl-young:
Just adding this comment to document the original issue:
In some state methods, state level primary data may need to be allocated based on national data. However it seems that if the source data is `state`, but the allocation dataset is `national`, the primary data gets aggregated to the national level.

catherinebirney:
> so the state water method isn't affected here?

@bl-young No state water is not impacted. The method changes how national water is attributed for some activities - the results shouldn't change. The old method aggregated the geoscale of the primary FBA first, then was multiplied by the ratios of the allocation df. This modified method multiplies first, then aggregates. But it is now more flexible.

catherinebirney:
> FYI, the approach (to cases where the attribution dataset is more aggregated than the source dataset) in the `seea-recursive` branch is very similar to what @catherinebirney has here.

Great - that was the impression I got glancing at your branch. We can pull my changes in now to help with state modeling efforts, but will ultimately use your code.

bl-young:
`Error: Process completed with exit code 137.` seems to be a memory issue error.
I was running into similar challenges with the CAP_HAP_state method I think just given the size of the dataframes when tracking things at the state or county level.

catherinebirney:
@bl-young @a-w-beck The changes to the github actions (mainly implemented by Andy) address the memory issues. Right now the gh actions are set to continue running even if an FBS fails, which we can change back so they fail. CAP_HAP is currently failing because of the recent changes to the MECS FBA. The FBA on data commons does not have table number in the description - there is currently an issue with Data Commons where I cannot upload any files. I will upload a revised MECS FBA before we pull this PR into develop once I can access DC again.

catherinebirney:
> @bl-young @a-w-beck The changes to the github actions (mainly implemented by Andy) address the memory issues. Right now the gh actions are set to continue running even if an FBS fails, which we can change back so they fail. CAP_HAP is currently failing because of the recent changes to the MECS FBA. The FBA on data commons does not have table number in the description - there is currently an issue with Data Commons where I cannot upload any files. I will upload a revised MECS FBA before we pull this PR into develop once I can access DC again.

Ah actually, I will check if updating the MECS FBA on data commons will break the FBS on the master branch

catherinebirney:
@bl-young I didn't get the gh actions testing quite right - the gh action still fails at the first FBS failure rather than running through all FBS and reporting which FBS failed, but i think we should go ahead and merge this PR and continue updating the gh actions later as this PR does fix the memory issue impacting most of the branches

bl-young:
Works for me

catherinebirney:
changes to geoscale impacts results of land FBS and water m3 FBS - but changes are less than 0.2% for all data rows (water m3 rounds to 0)

PR #248, opened 2022-07-12
Add 2018/2019 state employment FBS

catherinebirney:
- also add 2019 to BLS QCEW FBA list


bl-young:
Data Commons I expect would be easier for Mo. That sounds good to me.

I noticed there is an old 2019 BLS FBA dataset up there (from v0.0.2).
https://edap-ord-data-commons.s3.amazonaws.com/flowsa/FlowByActivity/BLS_QCEW_2019_v0.0.2_5d6373b.parquet
Don't know if that's going to cause issues for generating the 2019 employment FBS remotely.

catherinebirney:
good catch - I'll remove v0.0.2 

I can go ahead and pull these changes into develop/master and do a new release. Is there anything else that should be included in that new release? The pandas updates?

PR #270, opened 2022-09-07
Data vis

catherinebirney:
- adds additional column to FBS dfs "AttributionSources" which contains source name of primary attribution df, this column is required for graphing barcharts
- New data visualization functions: Sankey, stackedBarChart
- new VisualizationEssentials.csv with standard colors for graphing sectors
- new csv "sector_2012_names" which includes additional names beyond official NAICS
- in sector_aggregation() add option to return all possible combinations of sector lengths (results in a df with double counting),  used for graphing and add option to exclude a list of sectors from aggregation, sectors are not aggregated beyond the sector level provided, also used for graphing
- requires plotly, requires kaleido


catherinebirney:
@bl-young this branch is ready to be pulled in unless you have any comments about the graphs. I pulled the GHG FBS into this branch so I could test the barcharts (had to re-run the FBS to get the AllocationSources column). I included two example ghg plots, where I picked a few NAICS6 so the results are easier to see.

No impact category:
flowsaBarChart

Impact category:
flowsaBarChartImpactCat

catherinebirney:
Editing my last comment - need to run the FBS comparison check before merge, running now

flowsa/flowbyfunctions.py
```diff
@@ -1230,3 +1312,30 @@ def aggregate_and_subset_for_target_sectors(df, method):
     df_subset = subset_df_by_sector_list(df_agg, sector_list)
 
     return df_subset
+
+
+def add_column_of_allocation_sources(df, attr):
```
bl-young, line 1317: I recall discussing that we only really need the primary allocation source and not the helper source in this column

bl-young:
Would it be worthwhile to add kaliedo to requirements and to add `generate_data_visualization.py` to the install testing?

flowsa/flowbyfunctions.py
catherinebirney: Updated function to only include primary attribution source

catherinebirney:
> Would it be worthwhile to add kaliedo to requirements and to add `generate_data_visualization.py` to the install testing?

added

catherinebirney:
> > Would it be worthwhile to add kaliedo to requirements and to add `generate_data_visualization.py` to the install testing?
> 
> added

kaleido 0.1.0.post1 was only [released for windows](https://github.com/plotly/Kaleido/issues/61#issuecomment-740529914) because it fixed a windows-related issue in the v0.1.0 release. So that version is failing on the mac and Linux installs. I'm going to remove it as a requirement.

bl-young:
> kaleido 0.1.0.post1 was only [released for windows](https://github.com/plotly/Kaleido/issues/61#issuecomment-740529914) because it fixed a windows-related issue in the v0.1.0 release. So that version is failing on the mac and Linux installs. I'm going to remove it as a requirement.

I believe we can specify different requirements for different os though would need to remember how to do that. Remind me, later versions of kaleido don't work as far as you know (at least for Windows?). And Kaleido is only for sankeys?

catherinebirney:
> I believe we can specify different requirements for different os though would need to remember how to do that. Remind me, later versions of kaleido don't work as far as you know (at least for Windows?). And Kaleido is only for sankeys?

Good idea - I implemented the different package requirements and the checks are now successful. yes, later versions do not work (at least for windows) and it is only for sankeys

PR #274, opened 2022-09-21
add new energy data - EIA SEDS, AEO, update MECS

catherinebirney:
@bl-young Are these tables good to be pulled in?


bl-young:
I'll have to confirm I don't recall when I last ran them.

bl-young:
I'd like to review the final FBA one more time but I think these are ready

catherinebirney:
@bl-young your update to specify which table to pull in to allocate nat gas is what I would have proposed

bl-young:
> @bl-young your update to specify which table to pull in to allocate nat gas is what I would have proposed

So it turns out that we can't filter by the description FBA field as it gets dropped during cleaning [here](https://github.com/USEPA/flowsa/blob/energy_data/flowsa/dataclean.py#L30-L31). We will need a different approach.

catherinebirney:
@bl-young I addressed that issue in the waste branch to keep the description because I wanted to filter by the REI description - let me find that change and might be able to cherry pick that commit into the energy branch

catherinebirney:
@bl-young I cherry picked the commit to retain the description in FBS methods but did not test in the energy branch. In the FBS yaml there is now an "allocation_selection_fields" (like Matthew set up in his branch) where you specify a column and a list of info to keep. See the [REI_waste_national_2012.yaml](https://github.com/USEPA/flowsa/blob/waste/flowsa/methods/flowbysectormethods/REI_waste_national_2012.yaml#L24)

bl-young (approved):

PR #277, opened 2022-10-04
add action to enable user triggered FBS comparison

bl-young:


bl-young:
@catherinebirney  I can't fully test this until its on master. I wonder if after this PR we could go ahead and update master from develop (~15 minor commits) before this latest batch of new PRs?

catherinebirney (approved):
@bl-young Sure, go ahead and pull develop into master with these changes

PR #283, opened 2022-10-26
update ghgi-recursive-yaml with develop

catherinebirney:
@matthewlchambers Pulling in the develop branch will address the github action failures so we can test if there is an issue with the methods

new additions
- adds method status for FBA files (flowsa/methods/method_status.yaml) - we can skip files we know will fail in GH action test without causing GH action to fail @matthewlchambers we can add BTS_Airlines.py so the GH actions will not fail
- new manually triggered FBS comparison GH action
- drop support of 3.7 in GH action tests
- updates BEA data to most recent data in useeior @matthewlchambers this data update will change your GHG FBS results
- Retains the 'Description' column longer so in the FBS method we can select by description column
- corrects the URL for FBS methods saved to metadata file
- adds zenodo authors, adds paper badges to readme
- new FBAs: AEO, QWI, SEDS, additional MECS data @matthewlchambers you will need to update your MECS Energy FBA call to specify the tables (listed in Description col), see CAP_HAP_national_201.yaml for example
- new FBS files for state employment

PR #284, opened 2022-10-26
update ghgi-recursive-yaml from a copy of develop branch

catherinebirney:
@matthewlchambers Pulling in the develop branch will address the github action failures so we can test if there is an issue with the methods

new additions
- adds method status for FBA files (flowsa/methods/method_status.yaml) - we can skip files we know will fail in GH action test without causing GH action to fail @matthewlchambers we can add BTS_Airlines.py so the GH actions will not fail
- new manually triggered FBS comparison GH action
- drop support of 3.7 in GH action tests
- updates BEA data to most recent data in useeior @matthewlchambers this data update will change your GHG FBS results
- Retains the 'Description' column longer so in the FBS method we can select by description column
- corrects the URL for FBS methods saved to metadata file
- adds zenodo authors, adds paper badges to readme
- new FBAs: AEO, QWI, SEDS, additional MECS data @matthewlchambers you will need to update your MECS Energy FBA call to specify the tables (listed in Description col), see CAP_HAP_national_201.yaml for example
- new FBS files for state employment

NOTE: delete "develop_copy" branch after merge


catherinebirney:
@bl-young @matthewlchambers replace the PR #283 with this PR so can resolve conflicts in CAP_HAP_national_2017 without pushing to develop branch

bl-young:
ok I resolved, just needed to make sure the following was included. I think this was a feature that also exists on this branch...
```
        allocation_selection_fields:
          Description:
            - "Table 2.2"
            - "Table 3.2"
 ```

catherinebirney:
Ah - no, Matthew's code needs to be updated so the Description column is not dropped early on, replicating this [commit](https://github.com/USEPA/flowsa/pull/284/commits/ff6bd183d1568fe3035993728777efe2c8a468d3)

bl-young:
Looks like the actions on this branch are still not quite right. I think we can probably go ahead and pull in and then finish updating there? or did you want to wait? Wasn't sure if you were waiting for a more thorough review from me or not.

PR #292, opened 2022-11-02
new employment fbs - 2012, 2014, 2015

catherinebirney:


catherinebirney:
@matthewlchambers @bl-young I added the Employment FBS to Data Commons for 2012, 2014, 2015. Let me know if I should add any additional years.

bl-young (approved):

catherinebirney:
The CAP_HAP FBS is failing because the EIA_MECS data on Data Commons does not contain table numbers in the Description column

bl-young:
> The CAP_HAP FBS is failing because the EIA_MECS data on Data Commons does not contain table numbers in the Description column

Yes that's coming up elsewhere too.. I guess those will get published with the next release since they are on develop right now.

catherinebirney:
> > The CAP_HAP FBS is failing because the EIA_MECS data on Data Commons does not contain table numbers in the Description column
> 
> Yes that's coming up elsewhere too.. I guess those will get published with the next release since they are on develop right now.

I added the revised EIA_MECS_Energy FBAs to Data Commons, so the CAP_HAP FBS error should be resolved

catherinebirney:
changes in the land and water m3 FBS results are expected - these changes are due to changes in how geoscale is handled

PR #293, opened 2022-11-02
Add waste-related FBA and FBS

catherinebirney:


catherinebirney:
changes to FBS results
- CNHW_national_2014 - need to determine why results changed
- Land_national_2012 - differences expected, results impacted by changing geoscale method
- Water_national_2010_m2 - need to determine why results changed
- Water_national_2015_m2 - need to determine why results changed
- Water_national_2015_m3 - differences expected, results impacted by changing geoscale method

catherinebirney:
GH actions is failing with
Failed: Error generating: CAP_HAP_national_2017
but appears as FBS are all passing - need to revise GH actions before merge

bl-young:
> CNHW_national_2014 - need to determine why results changed

It looks like some changes you made including to estimate suppressed data in the BLS data before calculating tons per employee are likely the root cause https://github.com/USEPA/flowsa/commit/55574becdd2bcb7a03f78d684186665870558ebd

Looks reasonable to me. This will get updated again in the recursive branch when we shift to use the employment FBS.

bl-young:
Turns out all the stewi related FBS were failing. fixed by removing `v` as a passed parameter which had been added and then removed (f0c3c4f)

bl-young:
All methods now building, only these differences remain to resolve:

> * CNHW_national_2014 - need to determine why results changed
> * Land_national_2012 - differences expected, results impacted by changing geoscale method
> * Water_national_2010_m2 - need to determine why results changed
> * Water_national_2015_m2 - need to determine why results changed
> * Water_national_2015_m3 - differences expected, results impacted by changing geoscale method

PR #294, opened 2022-11-03
allow cached datasets to undergo futher selection

bl-young:
Proposal to allow cached datasets to be further filtered for a specific activity set. Should not impact existing methods with cached datasets.

See following simplified example:

```
sources_to_cache:
  EIA_MECS_Energy:
    year: 2014
    selection_fields:
      Class: Energy
      Unit: Trillion Btu
      Description:
          - Table 2.2
          - Table 3.2
    estimate_suppressed: !script_function:temp_data_source_functions estimate_suppressed_mecs_energy
    clean_fba: !script_function:temp_data_source_functions clean_mecs_energy_fba
    clean_fba_w_sec: !script_function:temp_data_source_functions clean_mapped_mecs_energy_fba
    attribution_method: proportional
    attribution_source:
      Employment_national_2018: # replace with 2014 when available
        data_format: FBS
        year: 2018

source_names:
  EPA_NEI_Nonpoint:
    geoscale: state
    year: 2017
    activity_to_sector_mapping: SCC
    fedefl_mapping: NEI
    activity_sets:
      industry_combustion_coal:
        selection_fields:
          PrimaryActivity: !from_index:NEI_Nonpoint_2017_asets.csv industry_combustion_coal
        attribution_method: proportional
        attribution_source:
          EIA_MECS_Energy:
            selection_fields:
              Flowable: Coal
```

### DIFF
--- a/flowsa/flowby.py
+++ b/flowsa/flowby.py
@@ -1003,11 +1003,22 @@ class FlowByActivity(_FlowBy):
 
         if attribution_method == 'proportional':
             # First check cache for FlowBy
-            if isinstance(fba.config['attribution_source'], str):
-                attribution_fbs = (fba.config['cache']
-                                   [fba.config['attribution_source']])
+            attribution_source = fba.config['attribution_source']
+            if isinstance(attribution_source, str):
+                # attribution_source as string MUST BE cached
+                attribution_fbs = (fba.config['cache'][attribution_source])
             else:
                 (name, config), = fba.config['attribution_source'].items()
+            if name in fba.config['cache'].keys():
+                # cached attribution_source passed as dictionary can include
+                # additional selection fields
+                attribution_fbs = fba.config['cache'][name]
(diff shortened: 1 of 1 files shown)


bl-young:
That makes sense to me - so that could include any additional data suppression or cleaning or something like that. I guess the limitation for the user is that you can only _add_ but not take away (which makes sense...). I can't think of any use cases right now but don't see a reason not to include the function.

FYI I expect this would be quite handy for use of EIA_MECS in the ghg methods as well.

matthewlchambers:
> That makes sense to me - so that could include any additional data suppression or cleaning or something like that. I guess the limitation for the user is that you can only _add_ but not take away (which makes sense...). I can't think of any use cases right now but don't see a reason not to include the function.
> 
> FYI I expect this would be quite handy for use of EIA_MECS in the ghg methods as well.

I expect so as well. I'll go ahead and merge this then.

PR #295, opened 2022-11-04
Add argument to FB constructor to prevent (when necessary) adding columns

matthewlchambers:
In general, when a `FlowBy` constructor is called, any missing columns are added. However, `_FlowBy.astype()` necessarily calls the constructor but is used in places where it is undesirable to add columns. The PR adds an argument `add_missing_columns` to the constructor. The default value for this argument is `True`, but when necessary, as in `asype()`, it can be set to `False`, in which case no new columns will be added.

### DIFF
--- a/flowsa/flowby.py
+++ b/flowsa/flowby.py
@@ -83,6 +83,7 @@ class _FlowBy(pd.DataFrame):
         self,
         data: pd.DataFrame or '_FlowBy' = None,
         *args,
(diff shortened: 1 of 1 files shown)


bl-young:
I may not get a chance to test this specifically soon but thanks for solving this! I'd say go ahead and pull in if you feel confident in the fix.

bl-young (approved):

PR #308, opened 2022-12-20
create FBA Blackhurst_IO with pdf in externaldata directory

catherinebirney:
add direct/indirect wat withdrawals for industrial sectors as pdf - cannot scrape data from pubs.acs.org


bl-young:
I was able to generate and retrieve this FBA, length of 431 records. I'm not familiar with the datasource but looks ok to me.

bl-young (approved):

catherinebirney:
@bl-young There are some unexpected changes to FBS results in this PR and in the stateior updates, see this [GH action](https://github.com/USEPA/flowsa/actions/runs/3741940461). I'm working through why there are differences. I assume there are changes to FBAs from v.1.2.4 to v1.3. The GH actions is mainly using v.12.4 FBAs on data commons while I generated all new FBAs locally when creating the v1.3 FBS

catherinebirney:
@bl-young The results for `CNHWC_national_2014` changed because the CDDpath FBA changed to include activityconsumedby, so now CNHWC has data for sectorconsumedby (v1.2.4 was "None"). Are you ok with this change?

There is an issue with `REI_waste_national_2012` because one of the activities contains a single quote in the string `Drugs and druggists’ sundries`. When I map REI to sectors locally, those rows of data are dropped. When mapped in github actions, those activities are retained. I'm not sure why I am losing the data locally when GH actions doesn't. I know we've run into this issue before with activity strings - do you remember how we've handled it?

bl-young:
> The results for `CNHWC_national_2014` changed because the CDDpath FBA changed to include activityconsumedby, so now CNHWC has data for sectorconsumedby (v1.2.4 was "None"). Are you ok with this change?

Yes that should be fine. We can confirm it still works next time it goes into a USEEIO update. As long as the flow quantities are the same and we've just added more detail (which I think is the case)

catherinebirney:
@bl-young There is a 1% increase in total flowamount - so I'm good with the changes

bl-young:
I think I"ve identified the issue - can we hold off on posting 1.3.0 of CNHWC? and I'll add an issue to resolve at a later point. I dont want to hold up this PR.

PR #315, opened 2023-01-26
V1.3.1 release

catherinebirney:
- Adds GHGI FBA and GHG_national FBS for 2020
- Updates FBS for 2016, 2017, 2018, and 2019 with latest inventory
- updates GHG FBS m1 to equally allocate BEA and EIA MECS to sectors rather than use employment for attribution and to use single FBS yaml file as basis for all years
- update some GHG FBS attribution source data years (MECS)
- update stackedbarchart() to use colors defined in visualizationessentials.csv, to include option to specify target sector level and to generalize attribution methods (direct vs attributed)
- updates to GHGI activity names


bl-young:
did you want to try to drop these functions in somewhere as well: https://github.com/USEPA/GHGbyindustrydocs/blob/main/workdocs/generateSummaryTable.py

catherinebirney:
> did you want to try to drop these functions in somewhere as well: https://github.com/USEPA/GHGbyindustrydocs/blob/main/workdocs/generateSummaryTable.py
@bl-young yes, I'll add those. Thanks!

WesIngwersen (changes requested):
The GHG FBS may still have issues. Will post one related issue.

WesIngwersen (approved):

flowsa/methods/flowbysectormethods/GHG_national_m1.yaml
```diff
@@ -0,0 +1,710 @@
...
+        - "Urea Fertilization" #CO2
+        - "Urea Consumption for Non-Agricultural Purposes" #CO2
+        clean_parameter: {"325310": 'ActivityProducedBy'} # Fertilizers
+      carbonates:
+        <<: *use_allocation
+        names:
+        - "Other Process Uses of Carbonates" #CO2
+        clean_parameter: {"325180": 'ActivityProducedBy'} # Other Basic Inorganic Chemicals
+      lead:
+        <<: *lead_allocation
```
MoLi7, line 193: @catherinebirney or @bl-young , could you confirm "Lead Production" in Table 2-1 is not used here?

bl-young: no, it is used, see line 71 which is where the name is defined (via Table 2-1)

I think the way it's structured currently in the yaml is not ideal and so I will likely move lines 70-71 down into this section here.

MoLi7: I see, thanks Ben!

I want to completely decompose this config file to be able to generate FBS for each source_name after line 126, e.g. Table 2-1 Rice Cultivation, Table 2-1 Coal Mining, etc. I'm planning to go through every `name` after line 126, comment (toggle off) all other names, and save a yaml for that name only. But since a few names (lead, cropland_allocation, etc) are reusing definitions before line 126, should I also toggle off the definition if I want it excluded those names?

bl-young: does d6b8561 help?

Yes if you want to do each individual activity by individually I think that would be what you'd have to do.

MoLi7: Yes, very helpful. Thanks Ben!

bl-young:
@catherinebirney with the release of the supply chain factors should we go ahead and pull this in?

catherinebirney:
yes, I'll merge and create a new release

PR #323, opened 2023-02-20
Fixes to improve ghgi mappings

bl-young:
- Fixes overallocation of CO2 to buses ([GHG-44](https://epalca.atlassian.net/browse/GHG-44))
- Fixes missing allocation to gov't sectors for commercial emissions (see [comment here](https://epalca.atlassian.net/browse/GHG-36?focusedCommentId=16116)), and therefore overallocation to all other sectors


bl-young:
@WesIngwersen feel free to pull in, or wait depending on your workflow

WesIngwersen (approved):

flowsa/fbs_allocation.py
```diff
@@ -417,15 +417,17 @@ def allocation_helper(df_w_sector, attr, method, v, download_FBA_if_missing):
...
-            np.where(modified_fba_allocation[sector_col_to_merge].isin(bea),
-                     modified_fba_allocation[sector_col_to_merge],
-                     modified_fba_allocation['Sector'])
-        modified_fba_allocation['HelperFlow'] = \
-            np.where(modified_fba_allocation[sector_col_to_merge].isin(bea),
-                     modified_fba_allocation['FlowAmount'],
-                     modified_fba_allocation['HelperFlow'])
+        # merge is in the bea list to prevent dropped data,
+        # unless these bea codes already are in the helper dataframe, then skip
+        if not modified_fba_allocation['Sector'].isin(bea).any():
```
WesIngwersen, line 422: @catherinebirney  FYI flagging you since this is a change in what i suppose is a widely use function

WesIngwersen (commented):
On second look, the transportation allocation doesn't align perfectly with the NGIAM. 
In the NGIAM, there is a lot more customization for certain fuel types of what they are allocated to.

flowsa/fbs_allocation.py
bl-young: Yes thanks meant to do that too - should be ok as I don't believe we use BEA as a helper allocation anywhere else (this chunk mostly protects FBS that allocate to BEA gov't codes and then have a helper allocation source like BLS). But Catherine should confirm

bl-young:
> On second look, the transportation allocation doesn't align perfectly with the NGIAM. In the NGIAM, there is a lot more customization for certain fuel types of what they are allocated to. the crosswalk has them more (but not perfectly) uniform right now. For instance, jet fuel for gov sectors is exclusively allocated to Fed Defense in the NGIAM, but in the crosswalk its spread to all 92* sectors. That is just one example.
> 
> I would recommend closely alignment for m1 unless you think it should be handled differently.

I will look again. The original flow mapping of the fuel type `Jet Fuel Transportation` is a superset of the ultimate mapping, which also depends on the A-79 vehicle type. So for example military aircraft are only mapped to 928110

PR #326, opened 2023-02-23
Simplify HFC handling from 4-102

bl-young:
Uses more granular data in latest GHGI (4-102) to:
- eliminate the need to subtract transportation emissions (instead can use more refined activity names)
- use more refined activity names to directly attribute commercial, industrial, and residential emissions without further cleaning
- simplify the allocation of foams through the combined used of urethane and polyurethane (requires minor updates to `subset_BEA_table()`

see [GHG-55](https://epalca.atlassian.net/browse/GHG-55)


WesIngwersen (commented):
The mapping of "Commercial" and "Industrial" activities is not consistent across tables from the GHGI.  In tables 3-8 and A-14 "Commercial" maps to  

2213
42
44
45
493
51
52
53
54
55
56
61
62
71
72
81

I prefer being consistent with the interpretation of Industrial and commercial across the tables, unless you think there is strong evidence not to

WesIngwersen:
Re: the  handling of "Other" HFCs and PFCs, I think capturing them but leaving them unmapped is OK at this point. Instead of 'Others' in the FBA though I wonder if we should rename this "HFCs and PFCs, unspecified"

WesIngwersen:
> The mapping of "Commercial" and "Industrial" activities is not consistent across tables from the GHGI. In tables 3-8 and A-14 "Commercial" maps to
> 
> 2213 42 44 45 493 51 52 53 54 55 56 61 62 71 72 81
> 
> I prefer being consistent with the interpretation of Industrial and commercial across the tables, unless you think there is strong evidence not to

I'm will make this change in another branch and PR off here

bl-young:
> The mapping of "Commercial" and "Industrial" activities is not consistent across tables from the GHGI. In tables 3-8 and A-14 "Commercial" maps to
> 
> 2213 42 44 45 493 51 52 53 54 55 56 61 62 71 72 81
> 
> I prefer being consistent with the interpretation of Industrial and commercial across the tables, unless you think there is strong evidence not to

No that's fine. I think also need to include the three gov't sectors: 927110, 923110, 923120 for commercial.

PR #329, opened 2023-03-02
Refac github actions

catherinebirney:
Fail FBS test after running all FBS, not at the first failed FBS

* testing with PR to develop, but will PR into v1.3.1-release branch


catherinebirney:
closing because I can manually run a workflow_dispatch event to troubleshoot

PR #342, opened 2023-04-26
update industry_spec_key() so it is not based on sector string length

catherinebirney:
- Update how target_naics are created so it is not based on the sector string length, instead determine target_naics using df of sectors by sector_length
- addresses error mapping to special sectors (e.g. F010) for summary models in [sector_aggregation](https://github.com/USEPA/flowsa/blob/b20e3f381d67e0a5b71f4832a645d4dea72d3c59/flowsa/flowby.py#L1997-L2000)
-  also reformat industry_spec dictionary so key is target naics length and sectors/values are in a list

### DIFF
--- a/flowsa/flowby.py
+++ b/flowsa/flowby.py
@@ -2046,21 +2046,13 @@ class FlowBySector(_FlowBy):
         :return:
         """
         naics_key = naics.industry_spec_key(self.config['industry_spec'])
-        # TODO fix error here causing special sectors (e.g. F010) to drop
-        # when doing a summary model because lenght does not align (target_naics)
-        # does not have the right number of digits
-
-        # subset naics to those where the source_naics string length is longer
-        # than target_naics
-        naics_key_sub = naics_key.query(
-            'source_naics.str.len() >= target_naics.str.len()')
 
         fbs = self
         for direction in ['ProducedBy', 'ConsumedBy']:
             fbs = (
                 fbs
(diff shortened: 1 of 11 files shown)


bl-young (approved):
> addresses error mapping to special sectors (e.g. F010) for summary models in [sector_aggregation](https://github.com/USEPA/flowsa/blob/b20e3f381d67e0a5b71f4832a645d4dea72d3c59/flowsa/flowby.py#L1997-L2000)

Confirming that this is functioning properly for my state summary models

matthewlchambers:
> @matthewlchambers in my update of mapping source_naics to target_naics I did include the option of including non_naics (e.g., "non_naics: ACC") but I didn't quite understand what was happening originally so I'm not sure I included it correctly and I'd appreciate if you would check that the non_naics map correctly

Yep, I'll check that! I like this, it's much smoother than what I had before!

matthewlchambers:
I really love the new crosswalk, and the use of `melt`! This is sooo much easier to follow than what I had before, in addition to being more robust.

I just pushed 3 commits to this branch, making the following minor changes:
1. I changed the code that loops through `industry_spec` and updates `target_naics` to make the logic a little simpler.
2. I updated the code for `non_naics` to make it clearer than my original version and (incidentally) allow the user to specify a list of non-NAICS codes if needed.

matthewlchambers (approved):

catherinebirney:
Looks good, thanks @bl-young and @matthewlchambers. I will merge.

PR #343, opened 2023-04-26
expand exclusion_fields

catherinebirney:
revise exclusion_fields to subset a df based on multiple columns

### DIFF
--- a/flowsa/data_source_scripts/USDA_CoA_Cropland.py
+++ b/flowsa/data_source_scripts/USDA_CoA_Cropland.py
@@ -247,30 +247,6 @@ def coa_cropland_parse(*, df_list, year, **_):
     return df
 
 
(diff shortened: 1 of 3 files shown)


bl-young:
I went to apply this to the NEI activity where we drop a list of flows from a list of activities. This required tweaking your query to work for lists using `in` instead of `=`. (of course I spent a long time trying to figure out why it wasn't working only to realize I had copied over your `ActivityConsumedBy` when I needed `ActivityProducedBy` 😭 ). In the end I realized I don't need that conditional since the activity set itself already limits the specific activities...

Either way the update appears to still work for your Land FBS, but makes it a bit more flexible for lists.

bl-young (approved):

catherinebirney:
Ha, what a journey you took! Good changes, thanks @bl-young!

PR #344, opened 2023-04-27
revise generateflowbysector() to enable access to already attributed FBA sources

catherinebirney:
Enable ability in FBS methods to access already attributed FBA source data. 

Necessary for Land FBS method for [urban activity set](https://github.com/USEPA/flowsa/blob/43024b45aed72c17cd949d008d4513e6aa5451fb/flowsa/methods/flowbysectormethods/Land_national_2012.yaml#L137-L143), need to access the already attributed data [CBECS](https://github.com/USEPA/flowsa/blob/43024b45aed72c17cd949d008d4513e6aa5451fb/flowsa/methods/flowbysectormethods/Land_national_2012.yaml#L13-L28) and [MECS](https://github.com/USEPA/flowsa/blob/43024b45aed72c17cd949d008d4513e6aa5451fb/flowsa/methods/flowbysectormethods/Land_national_2012.yaml#L29-L46)

- revise generateflowbysector() to include the already attributed data in the method config.

### DIFF
--- a/flowsa/data_source_scripts/USDA_ERS_MLU.py
+++ b/flowsa/data_source_scripts/USDA_ERS_MLU.py
@@ -22,6 +22,7 @@ from flowsa.literature_values import \
     get_urban_land_use_for_airports, \
     get_urban_land_use_for_railroads, get_open_space_fraction_of_urban_area
 from flowsa.validation import compare_df_units
+from flowsa.flowby import FlowByActivity
 
 
 def mlu_call(*, resp, **_):
@@ -83,8 +84,8 @@ def mlu_parse(*, df_list, source, year, **_):
     return dfm3
 
 
-def allocate_usda_ers_mlu_land_in_urban_areas(*, flow_subset_mapped,
-                                              fbs_list, **_):
+def allocate_usda_ers_mlu_land_in_urban_areas(
+        fba: FlowByActivity, **_) -> FlowByActivity:
     """
     This function is used to allocate the USDA_ERS_MLU activity 'land in
     urban areas' to NAICS 2012 sectors. Allocation is dependent on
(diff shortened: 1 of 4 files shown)


catherinebirney:
@matthewlchambers Do you have any comments on these changes? Or a better method for accessing the already attributed source data?

matthewlchambers:
I think a better way to access the already attributed data would just be to add the dataset that will need to be accessed to `sources_to_cache`. Then they can be accessed from `config['sources_to_cache']['...']`.

catherinebirney:
That is a better idea, thanks!

catherinebirney:
@matthewlchambers generateFlowBySector() still requires revision because the MECS and CBECS data are used as primary sources, not attribution sources

matthewlchambers:
> @matthewlchambers generateFlowBySector() still requires revision because the MECS and CBECS data are used as primary sources, not attribution sources

Is there a reason they can't just be listed both under `sources_to_cache` and as a primary sources (possibly using yaml anchors/aliases to reduce the amount of duplicate text? A concern I have is that with the code as written in the latest commit for this PR, if a source is cached, it's then impossible to use the same source differently (e.g. with different selection fields, etc.) as a primary source. Basically, the cache is overriding anything else with the same name.

catherinebirney:
If the sources are listed in `sources_to_cache` and as primary sources then the code is run twice, which I was trying to avoid. But I do understand your point of the cache overriding all other calls to the same data source, which is more problematic than code being run twice, especially as in our current version of flowsa, we consistently run the same code multiple times in an FBS method. So - I agree, I'll revert those commits and use anchors in the land FBS

catherinebirney:
Closing/not merging this PR because opted to not change flowby.py. Instead updated the Land FBS yaml in recursive-refac branch.

PR #347, opened 2023-05-03
Drop deprecated code and revise log handling

bl-young:
- old code for `flowbysector.py` is dropped
- deprecated functions are removed (primarily in `validation.py`, `dataclean.py`, `fbs_allocation.py`)
- data_source_script cleaning functions which now use more general approaches have been removed (e.g., `BEA.py`
- `temp_data_source_functions.py` have been moved to appropriate scripts
- logging functions are moved to `flowsa_log.py` (from `settings.py`) and duplicate code removed
   - log files are enabled for FBA and FBS (to be refined in future)


flowsa/dataclean.py
```diff
@@ -231,91 +230,3 @@ def standardize_units(df):
...
-    df = df.assign(Spread=None)
-    # reset min, as current values are misleading
-    log.info('Reset Min to None')
-    df = df.assign(Min=None)
-    # reset min, as current values are misleading
-    log.info('Reset Max to None')
-    df = df.assign(Max=None)
-    # reset DR, as current values are misleading
-    log.info('Reset DataReliability to None')
-    df = df.assign(DataReliability=None)
```
catherinebirney, line 307: @bl-young Do we still want to reset the DQ scores to None?

bl-young: We can certainly add that back in to `flowby.py` - it never made it in there. I can do that.

bl-young: ok completed with 82af6df

catherinebirney (approved):
Thanks looks good, this can be pulled in

PR #349, opened 2023-05-10
minor updates to support CAP_HAP FBS

bl-young:
- final update to enable allocation of EIA MECS data from census regions to states for use as state level allocation source
- adds a common CAP_HAP method file to consolidate across methods
- new methods:
  - CAP_HAP_national_2020
  - CAP_HAP_national_m2 (w/ secondary contexts)


bl-young:
@catherinebirney - more of an FYI really...

catherinebirney (approved):
Thanks Ben - I'll apply the MECS attribution to the state land FBS

PR #354, opened 2023-07-14
update recursive-refac with proportional_attribution

catherinebirney:
- Major refactoring of attribution fxns - ability to have multiple non-recursive attribution methods, option to specify what columns to use to attribute primary data source with secondary data source, can split an FBS used as an attribution source into activity sets
- New attribution methods: multiplication, division, inheritence
- New attribution cleaning fxns: weighed_average, substitute_nonexistent_values
- new fxn to attribute national data to state data using employment
- new fxn to calculate flow per employee
- require use of "activity_schema" into source_config.yaml to determine if "sector-like_activities"
- ability to specify different NAICS sector years for different years of a data source
- generalized fxn to estimate suppressed sectors using equal attribution
- delete obsolete code (and comment out other fxns)
- update seeAvailableFlowByModels() so does not return yaml files that end in "summary" or "common"


catherinebirney:
@bl-young This code still needs some tweaks, but I think it is ready to be pulled into the recursive-refac branch.  I did comment out most of the data vis fxns which need to be reworked

bl-young (approved):
I expect we'll do some more editing and testing but I think fine to pull in.

PR #357, opened 2023-08-03
Add action to create an FBA

bl-young:


bl-young:
@catherinebirney seems to work fine[ see example EIA_MECS 2018](https://github.com/USEPA/flowsa/actions/runs/5755311640)

PR #360, opened 2023-08-09
update methods to estimate suppressed data

catherinebirney:
- when estimating suppressed data, we now account for values of grandchildren
- new validation check to compare flow totals in a df before and after attribution
- updates to MLU functions to work for state methods
- update to substitute_nonexistent_values() to include all possible sectors at the state level
- drop 'AG LAND' mapping from USDA CoA Cropland Crosswalk because AG LAND (11) does not represent the summation of 111 and 112
- revisions to state water, state land, national land FBS yamls


bl-young:
I successfully tested GHG_national_2017 and TRI_DMR_state_2012. The former shows changes (expected due to use of employment FBS) while the latter does not.

I did not review in depth your adjustments.

PR #363, opened 2023-08-25
Split flowby.py into 3 python scripts

catherinebirney:
Move class-specific code out of flowby.py into 
- flowbyactivity.py (moved code in old flowbyactivity.py into generateflowbyactivity.py)
- flowbysector.py

Because flowby.py was very long and it was getting difficult to follow. 

@matthewlchambers @bl-young any reason we should not split up the code?


bl-young:
assuming it works I think this is a good idea

catherinebirney:
> assuming it works I think this is a good idea

:satisfied: yes, I will look into the flake8 failures

matthewlchambers:
I think this is a great idea. The one change I would make is to not create the `flowbyseries.py` file, but to keep the class `_FlowBySeries` in `flowby.py`, the class `_FBASeries` in `flowbyactivity.py`, and the class `_FBSSeries` in `flowbyseries.py`. Just because those classes are crucial to the classes `_FlowBy`, `FlowByActivity`, and `FlowBySector` functioning.

catherinebirney:
> I think this is a great idea. The one change I would make is to not create the `flowbyseries.py` file, but to keep the class `_FlowBySeries` in `flowby.py`, the class `_FBASeries` in `flowbyactivity.py`, and the class `_FBSSeries` in `flowbyseries.py`. Just because those classes are crucial to the classes `_FlowBy`, `FlowByActivity`, and `FlowBySector` functioning.

Great input, thanks, I'll make that change!

catherinebirney:
I will pull this branch into recursive-refac once the checks on PR #364 are complete

PR #365, opened 2023-08-28
Updates to GHG state method

bl-young:


bl-young:
fyi @WesIngwersen these are the updates I was describing

catherinebirney (approved):
I did not run the state models, but the changes look good to me

bl-young:
pulled this into the other ghg_updates branch (for national updates) because needed some of the fixes there as well

PR #372, opened 2023-09-06
Bea refac

catherinebirney:
- rename all BEA FBAs to align with useeior naming schema
- adds BEA summary make and use tables for 2012-2020

@bl-young I changed the BEA naming schema so the csv files and FBA yaml files have the same name...which align with the useeior table names. I think I caught all uses of BEA FBAs, but it is possible a few have escaped me. 

I'm validating that the FBA results for the existing BEA tables have not changed


catherinebirney:
@bl-young This branch is ready to be pulled in if you are ok with the name changes

I changed the units to keep the original "Million USD" - previously when generating the FBAs, we were converting to USD

PR #375, opened 2023-09-08
update develop with changes to harmonize state and national fbs values

catherinebirney:
- Modify the water, land, employment, ghg FBS so state sectors sum to national sectors. required modifying both state and national fbs yamls
- new function compare_national_state_fbs() which compares aggregated results of state vs land FBS
- add annual BEA make and use tables
- change BEA FBA names to align with useeior naming schema
- Updates to BLM FBAs - adds previously dropped state and national data for select activities
- Add D.C. data to ERS MLU and NWIS WU
- modifications to allow for filling a primary data source Location col with Location col from attribution data
- move equally_attribute() to flowby.py and modify to work for both FBAs and FBS
- new Employment FBS state and national years


bl-young:
@catherinebirney you may have already seen this but looks like the links on the USGS site itself are bad for these ones that are failing

catherinebirney:
> @catherinebirney you may have already seen this but looks like the links on the USGS site itself are bad for these ones that are failing

Yes, thank you! I am working on those now. I plan on adding the MYB datasets to method_status and reaching out to USGS to see what is going on

PR #378, opened 2023-09-20
Revised approach to GHG models

bl-young:
## Major updates

- Update to use a summary FBS model as the primary source for state models and national detail models for GHG and CAP_HAP
- New Detail Supply (Make) and Use tables (#383) which are built from summary tables attributed to detail using the best available detail model. Requies 2017 BEA to 2017 NAICS schema and 2017 NAICS crosswalk.
- New Detail level NAICS target spec to mimic best available NAICS for a detail (BEA) model

## GHGs
**National_m1**: The new summary method. This uses new Detail Make and Use to better align data years.

**National_m2**: A new detail method that, where possible, starts from the summary level FBS (m1), before further attributing to the detail level. In other cases, or where direct attribution to 6 digits is required, m2 returns to the original FBA (GHGI) to attribute directly. Flow totals by sector for m2 should aggregate to those in m1.

**State_m1**: Summary level state model that uses as its primary source the national summary model, and attributes flows to state proportionally to the StateGHGI. In some cases where data granularity is not consistent with national model (i.e., not as granular), stateio use tables are used.

The old National detail method (old m1), and the draft state method attributing all data from the StateGHGI (old m1) are dropped. See https://github.com/USEPA/flowsa/commit/8d2daec496dd61730348f74336cf34463db9e4ad and https://github.com/USEPA/flowsa/commit/fb08a00d9b0d7ab8d43113394462714f54635092

## CAP-HAP
Detail level state models serve as the starting point for the CAP_HAP models. Due to size the Nonpoint, Nonroad, and Onroad FBS are modeled separately.

**State_m1**: Aggregates the three FBS to summary level.

**National_m1**: State detail components aggregated to national level.

##
v2.0 was tagged at [v2.0-GHG_methods](https://github.com/USEPA/flowsa/releases/tag/v2.0-GHG_methods) for the status of these methods prior to this change.


bl-young:
@WesIngwersen @catherinebirney I am generating all the methods now: national summary (m1), national detail (m2), and state (m1), and the `compare_national_state_fbs()` outputs

Once these are confirmed ok I will set this as ready for review.

bl-young:
@catherinebirney let me know what you think about the tag of the old methods and we can include that information here in the description of this PR.

catherinebirney:
> @catherinebirney let me know what you think about the tag of the old methods and we can include that information here in the description of this PR.

I'm good with either a tag of the develop or 2.0 release branch.

bl-young:
For some activities/sectors we may want to attribute using the state use table instead of the StateGHGI (e.g. airline transport)

bl-young:
> > @catherinebirney let me know what you think about the tag of the old methods and we can include that information here in the description of this PR.
> 
> I'm good with either a tag of the develop or 2.0 release branch.

ok tag implemented and referenced above

bl-young:
@catherinebirney if the remaining FBS testing passes with no new issues then I think this is ready to pull in. We are going to need a different approach for CAP_HAP 2014 Nonroad and I don't want that to hold things up.

catherinebirney:
> @catherinebirney if the remaining FBS testing passes with no new issues then I think this is ready to pull in. We are going to need a different approach for CAP_HAP 2014 Nonroad and I don't want that to hold things up.

Great! I tried running nonroad 2014 locally, but was not able to. I agree we should go ahead and pull in and address nonroad later

PR #382, opened 2023-09-22
Update flowsa2.0 with updates from reviewer comments

catherinebirney:
Revising code based on flowsa2.0 review by DG and JB

- move all fxn defintions out of _init_.py
- move test files into their own directory
- update text explanations
- in README example of FBA, use an FBA that does not require an API key


catherinebirney:
@bl-young The commits in this PR address the reviewer comments that should be addressed prior to the 2.0 release. I've added the remaining comments as tasks on Jira to be addressed/discussed at a later date. 

Once these changes are pulled in, we are ready for the 2.0 release, pending deciding how to address the GHG FBS

bl-young (approved):

PR #383, opened 2023-10-12
2017 Supply and Use tables from useeior

WesIngwersen:
- Adding 2017 Detail and Summary Supply and Use tables and updating script notes. Manually spot-checked csv table data against originals.
- Enables use of 2017 NAICS crosswalk and `target_naics_year`
- Adds FBS of annual summary use table attributed to detail sectors proportionally to Detail 2017 Use table


WesIngwersen:
fyi @catherinebirney  note that I pulled in and added the metadata json files from useeior for these two files... i found that would be very valuable to have here

catherinebirney:
> fyi @catherinebirney note that I pulled in and added the metadata json files from useeior for these two files... i found that would be very valuable to have here

I have wondered if it would be worth putting the csv files (and metadata) on the useeior data commons repo instead of within flowsa and have flowsa load the data from there - it could be a cleaner/easier way of tracking what files we use

bl-young:
> I have wondered if it would be worth putting the csv files (and metadata) on the useeior data commons repo instead of within flowsa and have flowsa load the data from there - it could be a cleaner/easier way of tracking what files we use

I think that's a nice idea. Since they are rda files they won't be impacted by the useeior package after they are created, so just using the metadata from useeior (as Wes did here) and saving them on data commons would be sufficient. We would not need these scripts anymore to document them.

bl-young:
Also I believe there is some conflict/duplication here with edits made to the scripts. Now there is both a `pull_BEA_Make_and_Use_csv.py` and `write_BEA_data_from_useeior.py`

catherinebirney:
> Also I believe there is some conflict/duplication here with edits made to the scripts. Now there is both a `pull_BEA_Make_and_Use_csv.py` and `write_BEA_data_from_useeior.py`

you can keep 'write_BEA_data_from_useeior' and delete 'pull_BEA_Make_and_Use_csv.py' - I was just trying to clean up the documentation yesterday

bl-young (approved):
I think for now we can pull this in while we consider @catherinebirney idea.

I will note that for generating the FBA's, the Supply tables are transposed compared to the old Make tables, so we will want to account for that in the script. Industries are on the columns

bl-young:
@catherinebirney do you want to work on the FBAs (and FBS?) here in this branch?

catherinebirney:
> @catherinebirney do you want to work on the FBAs (and FBS?) here in this branch?

@bl-young Sure - I will work on the FBAs now in this branch.

catherinebirney:
@bl-young I added the initial FBA code - I'm checking for any issues now. Do you want to generate the FBS?

bl-young:
> @bl-young I added the initial FBA code - I'm checking for any issues now. Do you want to generate the FBS?

Sure i will give it a shot

bl-young:
@catherinebirney draft BEA crosswalk added. It needs more review and will likely change, but includes 2012 and 2017 NAICS for now.

flowsa/common.py
```diff
@@ -77,7 +77,8 @@ def load_crosswalk(crosswalk_name):
                'sector_name': 'Sector_2012_Names',
                'household': 'Household_SectorCodes',
                'government': 'Government_SectorCodes',
-               'BEA': 'NAICS_to_BEA_Crosswalk'
+               'BEA': 'NAICS_to_BEA_Crosswalk',
+               'BEA_2017': 'NAICS_to_BEA_Crosswalk_2017',
                }
```
bl-young, line 82: @catherinebirney this was a temporary solution, I'm sure we can come up with something more robust

catherinebirney: I'm editing that fxn in my branch - I'm getting rid of that dictionary entirely

flowsa/flowby.py
```diff
@@ -673,7 +673,13 @@ def attribute_flows_to_sectors(
...
                 )
             elif self.config['data_format'] in ['FBS']:
-                fb = grouped.sector_aggregation()  # convert to proper industry spec.
+                # ensure sector year of loaded FBS matches target sector year
+                fb = naics.convert_naics_year(
+                    grouped,
+                    f"NAICS_{self.config['target_naics_year']}_Code",
+                    grouped['SectorSourceName'][0])
+                # convert to proper industry spec.
+                fb = fb.sector_aggregation()
```
catherinebirney, line 682: @bl-young This change should check the FBS sector year and convert to target year if necessary - none of my FBS requires conversion so I will need to validate once you integrate "Detail_Use_year" FBS into the ghg/cap_hap methods

bl-young: ok just pushed them. i didn't test yet on top of your commit

bl-young:
@catherinebirney I've reviewed the use tables, including converting them back to summary using `map_to_BEA_sectors()`, while they aren't a perfect match (due to many:many NAICS:BEA), I feel comfortable moving forward. Any thing left on your end before we pull this into the other branch?

bl-young:
> ~The Detail_Use_year tables do not get their NAICS codes converted to target year 2012- they are remaining in NAICS 2017. I'm working on addressing why~ Nevermind - the conversion happens later in the method. Let me finish running this FBS and then I am fine with pulling this into the state ghg branch. I'll comment again

Yeah it looks like its working fine in the CAP HAP methods

catherinebirney:
@bl-young The BEA_Use tables naics are now being converted correctly - I agree this PR can be merged

PR #386, opened 2023-10-18
update develop with Flowsa2.0 release branch

catherinebirney:
@bl-young I'd like to pull the release branch into develop and then into the state ghgi branch so we can test how all of the getflowby functions handle now that we've moved them out of the init.py script. Do you agree?


bl-young:
yep absolutely

PR #392, opened 2023-11-03
skip `select_by_fields` when using activity_sets

bl-young:
This avoids duplicating the `select_by_fields` step for `activity_sets` which can drop data when Activity or flow names are changed


bl-young:
@catherinebirney I think if we revert e220e8d on state_GHGI_v2, then we can squash and pull this in.

catherinebirney:
great, thanks Ben! That is a much better fix. I reverted the commit on state_ghgi_v2

PR #398, opened 2024-01-04
option to generate FBS that contains activity cols and sector name cols

catherinebirney:
generate an FBS that has columns of the original activity names and/or columns of the sector/naics names

can generate by running the example code

```
import flowsa.flowbysector
import pandas as pd

fbsmethod = 'Water_national_2015_m1'

fbs = pd.DataFrame(flowsa.flowbysector.FlowBySector.generateFlowBySector(
    fbsmethod,
    download_sources_ok=True,
    retain_activity_columns=True,
    append_sector_names=True
))
```

### DIFF
--- a/flowsa/common.py
+++ b/flowsa/common.py
@@ -73,8 +73,9 @@ def load_crosswalk(crosswalk_name):
     Used to load the crosswalks:
 
     'NAICS_Crosswalk_TimeSeries', 'NAICS_2012_Crosswalk',
-    'Sector_2012_Names', 'Household_SectorCodes', 'Government_SectorCodes',
-    'NAICS_to_BEA_Crosswalk_2012', 'NAICS_to_BEA_Crosswalk_2017'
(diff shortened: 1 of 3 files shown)


catherinebirney:
@bl-young do you think it would be better to implement these options through parameters in the FBS method yaml?

bl-young:
hmmm, I could go either way. The sector names seems less important to put in the yaml because it doesn't change the structure of the output (just adds a column), whereas keeping the activity name means that it could be much larger.

catherinebirney:
Thanks - I will leave as-is then. My thoughts are also that adding the activity and sector name columns are not part of the method - and I would not want there to be two method files just so on one of them we can append the activity names.

catherinebirney:
@wadedavis13 Let us know if these changes meet your requirements - if so, we can pull these changes into the develop branch

catherinebirney:
These changes are ready to be pulled into develop/master. If @wadedavis13 needs additional changes, the code can be modified after the next flowsa release

bl-young (approved):

PR #406, opened 2024-04-22
Updated GHG FBS

bl-young:
- Updates EPA_GHGI through 2022 (2012-2022)
- New BEA data for 2012-2022 (Summary & Gross Output), 2012 & 2017 Detail (using 2017 BEA schema)
- BEA_Detail FBS for 2013-2016, 2018-2022
- Updated GHG national FBS (m1 and m2) 2012-2022
- Includes new CoA data in #405


bl-young:
FBS for these models have been added to sharepoint. This PR will remain in draft until the final csv files are published by EPA.

WesIngwersen:
@bl-young Can we merge this now?

bl-young:
@WesIngwersen I've updated the validation after fixing an error in the 2012 FBS, see here: https://github.com/USEPA/GHGbyindustrydocs/commit/3112d40647715f7ab9c4ba9901a8aa3d9c4bb0bd.

We still have one minor issue in m2 for 2012-2016, where emissions from liming are getting dropped, because we use the Use table to attribute within crops sectors (`111`). And in the 2012 Use table, there is no use of the Lime sector by agricultural sectors (which has always been the approach).

The quantity of emissions is relatively small, in fact running the validation with with the summary models (where this issue is not present) barely changes the total flow validation check.

PR #444, opened 2025-04-15
add 2020 census data set for urban/rural splits

bl-young:


bl-young:
pending addition of 2000 dataset too

Issue #435, opened 2024-12-02
EPA GHGI change in zip file structure


bl-young:
The [2024 release](https://www.epa.gov/ghgemissions/inventory-us-greenhouse-gas-emissions-and-sinks-1990-2022) (2022 data year) of the GHGI has an updated zip file that is missing chapter 4.

bl-young:
The correct set of zip files is in an interior zip called `Main Text CSVs.zip`

![Image](https://github.com/user-attachments/assets/fd453ab2-125e-42d3-81db-33b1255efca2)

PR #451, opened 2025-06-04
Release v2.0.6

bl-young:
- Resolves error in accessing `EPA_GHGI` 2024 release (#435)
- Adds draft documentation guide
- Adds NAPCS crosswalk from Census

No new data objects published.


bl-young:
I'd like to do a release with #435 resolved prior to the next release with the updated GHGI data. cc @catherinebirney

WesIngwersen (changes requested):
NAICS_to_NAPCS_Crosswalk_2017.csv is not truly a crosswalk. Its a set of flows based on the Census 2017 survey without data. So maybe better not to include here in this release until we think more about use of this.

PR #457, opened 2025-09-03
V2.1 release

catherinebirney:
### Major changes:

1. Implemented ability to calculate and track data quality (DQ) scores for FBS
     - DQ scores for Data Collection, Data Reliability, Geographical Correlation, Technological Correlation, Temporal Correlation
     - DQ scores based on EPA's Guidance on Data Quality Assessment for Life Cycle Inventory Data
     - DQ scores are beta version in this release and not included in all FBS methods
     - All FBS methods include scores for Geographical Correlation, Technological Correlation, and Temporal Correlation
     - Full DQ implementation will require updating FBA methods with Data Collection and Data Reliability scores
     - New functions added to calculate DQ
          -	New `adjust_dqi_reliability_collection_scores()` to modify data reliability and data collection based on source and target sector levels
          -	`assign_temporal_correlation()` assigns temporal DQ based on difference between year of data and target year of FBS
          -	`assign_geographical_correlation()` assigns DQ for geoscale based on data geoscale vs target FBS geoscale
          -	`assign_technological_correlation()` assigns DQ scores based on difference between source and target sectors

1. Modified how activities are mapped to sectors to enable proper accounting for Technological Correlation scores, which are based on the difference between the original activity to sector mapping and the target sector level.
     -	First map all activities to the sector year identified in data crosswalk, then later convert to target sector year. Previously we immediately converted the crosswalk to target sector year, before matching on activities
     - We modified how NAICS are converted between NAICS years
          - We originally mapped all activities to NAICS6+ in the activity to sector crosswalk, then converted between NAICS years, then aggregated to target sector level, then merged NAICS to the activity-based data sets. This method is problematic when assigning DQ scores and unnecessary for FBS methods that are generated for more aggregated sector levels
          - Now to convert, we map the activities to the original sector year associated with that data. We then identify how many child NAICS there are (at NAICS 6) for each of the sectors and determine how many of those child sectors are converted to new sectors for the target sector year in `generate_naics_crosswalk_conversion_ratios()` and proportionally attribute the sectors to the new sectors for the target sector year.
               - For example, if we are converting NAICS4 across years, we identify all child NAICS6 for each NAICS4 and determine how those NAICS6 map between years. If there are five child NAICS6 and one child NAICS6 maps to a different parent NAICS4 in the target year, than 1/5 of the original NAICS4 parent value is mapped to a different NAICS4 in the target year
               - Conversion is not based on numeric values within the FBS because we might only have NAICS4 values, not NAICS6 and therefore do not have the data to create proportional conversions 
     -	New `subset_sector_key()`
              -	Subsets sector key to return sector/industry that most closely maps activity/source sectors to target sectors – drops parent sectors within crosswalk and assigns tech corr scoring, modifies DatarReliability and DataCollection scores based on mapping
     -	Modified NAICS conversion data check - originally checked if a sector-like activity was found in any NAICS year outside of the target year and if so, mapped to target year. This function did not always map correctly because the sector could be found in multiple NAICS years, and the NAICS years map differently to target year sectors
          - Revised function to check for the closest NAICS year to the target year and use that year to map to target NAICS

1. Updated default NAICS year in Employment FBS to NAICS 2017 (revised from NAICS 2012)
     - Changing NAICS year impacts the results of all FBS that use employment FBS as an allocation source
     - Previously, some BLS QCEW data were imported as NAICS 2012 left as NAICS 2012, while other data years were imported as NAICS 2017 and converted to NAICS 2012
     - Now, some BLS QCEW data are imported as NAICS 2012 and converted to NAICS 2017, while other data years are imported as NAICS 2017 and left as NAICS 2017
     - Many of these Employment datasets published as NAICS 2017 are later converted back to NAICS 2012 for use as allocation sources in other FBS methods. A conversion from NAICS 2012 -> NAICS 2017 -> NAICS 2012 occurs, which changes the employment results based on our conversion functions, resulting in changes to those relevant flows in the FBS methods. 

1. Modified how data are merged on location so we can correctly merge state with county data

### Minor changes:
- Correct error in `attribute_flows_to_sectors()`
     -	Original group_total assignment was based on original df FlowAmount values, but we reset the index, so needed to base group_total on new index of the df
- Adds FIPS scale (1,3,5) to FIPS_Crosswalk
- Add NAICS 2002, 2007, 2022 crosswalks
- Expand NAICS_Crosswalk_TimeSeries to include NAICS 2022
- New NAICS_Year_Concordance which maps published 6-digit sectors across years
- New Sector_Levels .csv which labels sector level and sector length for all sectors
- Update BLS_QCEW NAICS years for 2011, 2022, and 2023 in source_catalog.yaml
- BLS QCEW `estimate_suppressed_qcew()`
     -	Update the function to only estimate suppressed data up to max sector level. No longer estimate suppressed 6-digit sectors, when our target is 3-digit
- Consistent fips scale assignments. National = 5, state = 2, county = 1
- url updates to government FBA links
 
### FBA Changes
- Generates new FBAs for EPA GHGI for 2019-2023
- Updated BEA FBAs (Supply, Use, GrossOutput) for 2012-2023

### FBS Changes
- Updates to GHG FBS national (m1 and m2) for 2019 - 2023; drops 2012 - 2018 FBS which no longer will work with the latest FBAs
- New FBS method: Wages_national for 2017
- Updates Use and Supply tables in SUT format (see #453)

### Includes PR:
#441
#452 
#453 
#455 
#456


bl-young (approved):

bl-young:
Looks good to me - made a few minor changes to the PR description.

PR #458, opened 2025-10-28
modifications for updated pandas/numpy packages

catherinebirney:
- Code modifications required due to updating package requirements to to pandas 2.2+ and numpy 2.0+
- The changes to this code in this PR do not impact the results of FBA or FBS methods.
- Drop support for python 3.9, add support for 3.12
- Test github actions using python 3.12, updated from 3.10


bl-young:
Reviewing the FBS diff files:
- Looks like some diffs are showing up due to differences in how NaN's are handled: 
image
A minor issue that just makes the diff files longer than needed.


- Also flagging that StEWI based datasets are showing differences likely due to a more recent Facility matcher dataset being available on the Commons (v1.2.1), so despite the stewicombo data being static, how those facilities are assigned to sectors may have shifted some. In the future we may want to consider if allowing this shift is ok, or if we need to allow fixing the FRS file as well.

setup.py
bl-young: Due to numpy we need to bump this to >= 3.10, right?

bl-young (approved):

setup.py
catherinebirney: yes, addressed, thanks!

WesIngwersen:
I ran Employment_national_2022 successfully without any WARNINGs but was not able to compare it to a previously generated dataset

catherinebirney:
> I ran Employment_national_2022 successfully without any WARNINGs but was not able to compare it to a previously generated dataset

@WesIngwersen Your inability to compare FBS might have something to do with Ben's comment above about how handling nan has changed. I'm going to look into it this afternoon.

PR #461, opened 2026-01-06
Update acs census

elanphearerg:
I added in a new variable to be pulled from ACS and updated the units to correctly note where a value is a number, not a percentage


catherinebirney (approved):
looks good, you can go ahead and pull in!

